PR #5, opened 2020-12-30
Decode user and password from env configured proxy

leipert:
If someone sets an env variable defining a http_proxy, containing a
username / password with percent-encoded characters, then the resulting
base64 encoded auth header will be wrong.

For example, suppose a username is `Y\X` and the password is `R%S] ?X`.
Properly URL encoded the proxy url would be:

    http://Y%5CX:R%25S%5D%20%3FX@proxy.example:8000

The resulting proxy auth header should be: `WVxYOlIlU10gP1g=`, but the
getters defined by ruby StdLib `URI` return a username `Y%5CX` and
password `R%25S%5D%20%3FX`, resulting in `WSU1Q1g6UiUyNVMlNUQlMjAlM0ZY`.
As a result the proxy will deny the request.

Please note that this is my first contribution to the ruby ecosystem, to
standard lib especially and I am not a ruby developer.

Sorry for that and a happy and healthy 2021!

References:

- https://gitlab.com/gitlab-org/gitlab/-/issues/289836
- https://bugs.ruby-lang.org/projects/ruby-master/repository/trunk/revisions/58461
- https://bugs.ruby-lang.org/issues/17542

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -1180,7 +1180,8 @@ module Net   #:nodoc:
     # The username of the proxy server, if one is configured.
     def proxy_user
       if ENVIRONMENT_VARIABLE_IS_MULTIUSER_SAFE && @proxy_from_env
-        proxy_uri&.user
+        user = proxy_uri&.user
+        unescape(user) if user
       else
         @proxy_user
       end
@@ -1189,7 +1190,8 @@ module Net   #:nodoc:
     # The password of the proxy server, if one is configured.
     def proxy_pass
       if ENVIRONMENT_VARIABLE_IS_MULTIUSER_SAFE && @proxy_from_env
-        proxy_uri&.password
+        pass = proxy_uri&.password
+        unescape(pass) if pass
       else
         @proxy_pass
       end
@@ -1199,6 +1201,11 @@ module Net   #:nodoc:
     alias proxyport proxy_port      #:nodoc: obsolete
 
     private
+
+    def unescape(value)
+      require 'cgi/util'
+      CGI.unescape(value)
+    end
 
     # without proxy, obsolete
 
--- a/test/net/http/test_http.rb
+++ b/test/net/http/test_http.rb
@@ -188,6 +188,23 @@ class TestNetHTTP < Test::Unit::TestCase
     end
   end
 
+  def test_proxy_eh_ENV_with_urlencoded_user
+    TestNetHTTPUtils.clean_http_proxy_env do
+      ENV['http_proxy'] = 'http://Y%5CX:R%25S%5D%20%3FX@proxy.example:8000'
+
+      http = Net::HTTP.new 'hostname.example'
+
+      assert_equal true, http.proxy?
+      if Net::HTTP::ENVIRONMENT_VARIABLE_IS_MULTIUSER_SAFE
+        assert_equal "Y\\X", http.proxy_user
+        assert_equal "R%S] ?X", http.proxy_pass
+      else
+        assert_nil http.proxy_user
+        assert_nil http.proxy_pass
+      end
+    end
+  end
+
   def test_proxy_eh_ENV_none_set
     TestNetHTTPUtils.clean_http_proxy_env do
       assert_equal false, Net::HTTP.new('hostname.example').proxy?


leipert:
Switched to using `CGI.unescape` over `URI.decode_www_form_component` as per:

https://gitlab.com/gitlab-org/gitlab/-/merge_requests/52368#note_500425249

leipert:
@hsbt Could you review this contribution? Thank you in advance 🙇 

Have also created this issue in the bug tracker: https://bugs.ruby-lang.org/issues/17542

lib/net/http.rb
```diff
@@ -1180,7 +1180,8 @@ def proxy_port
     # The username of the proxy server, if one is configured.
     def proxy_user
       if ENVIRONMENT_VARIABLE_IS_MULTIUSER_SAFE && @proxy_from_env
-        proxy_uri&.user
+        user = proxy_uri&.user
+        CGI.unescape(user) unless user.nil?
```
nobu, line 1184: As cgi.rb isn't required here, it results in `uninitialized constant Net::HTTP::CGI` error.

nobu (changes requested):

lib/net/http.rb
nobu: And why `unless user.nil?` instead of `if user`?

leipert: @nobu Thanks for the review, changed it to `if user` and `if pass` respectively and added a `require 'cgi'` at the top.

mame: Maybe `require "cgi/escape"` is enough?

leipert: @mame / @nobu  Changed it to `require "cgi/escape"`!

leipert: Wouldn't it be `require "cgi/util"` though? And what is the difference between `CGI::unescape` and `CGI.unescape`

leipert: Changed it to `require "cgi/util"`. Hope that is alright.

nobu: As for method calls, `::` and `.` are equivalent.
`.` seems preferred in these days though.

And, as `CGI` is used only when proxy env is given, it can be deferred to that cases only.

leipert: Thank you @nobu. I don't understand the following aspect completely:

> And, as CGI is used only when proxy env is given, it can be deferred to that cases only.

I don't know how the Ruby dependency loading works exactly (I was wondering that the tests didn't catch `uninitialized constant Net::HTTP::CGI`) you described in the first place.

Should I use `autoload` for the CGI module (`autoload :CGI, 'cgi/util'`), or should I potentially require it someplace else, e.g.:

- https://github.com/ruby/net-http/blob/master/lib/net/http.rb#L650-L655

- https://github.com/ruby/net-http/blob/master/lib/net/http.rb#L1177-L1182

lib/net/http.rb
```diff
@@ -1180,7 +1181,8 @@ def proxy_port
     # The username of the proxy server, if one is configured.
     def proxy_user
       if ENVIRONMENT_VARIABLE_IS_MULTIUSER_SAFE && @proxy_from_env
-        proxy_uri&.user
+        user = proxy_uri&.user
+        CGI::unescape(user) if user
```
arturoherrero, line 1185: Related to https://github.com/ruby/net-http/pull/5#discussion_r571528379, I think they are referring to move the `require 'cgi/util'` L24 inside the conditions:
```suggestion
        user = proxy_uri&.user
        if user
          require 'cgi/util'
          CGI::unescape(user)
        end
```

lib/net/http.rb
leipert: @nobu Moved back to the `.` syntax and moved the require to the conditionals.

lib/net/http.rb
arturoherrero: An alternative to this is to extract a private method and re-use it in both places:
```ruby
unescape(user) if user

private

def unescape(string)
  require 'cgi/util'
  CGI::unescape(string)
end
```

nobu: > Related to [#5 (comment)](https://github.com/ruby/net-http/pull/5#discussion_r571528379), I think they are referring to move the `require 'cgi/util'` L24 inside the conditions:

I agree on this.

leipert:
@nobu / @mame I have moved the `unescape` and `require` to a private method. How do you feel about it?

jeremyevans (approved):
I'm in favor of this change, and the implementation seems fine.

nobu (approved):

leipert:
Thank you @nobu / @mame / @jeremyevans / @arturoherrero! 👍

Issue #6, opened 2021-01-13
Replacing Timeout.timeout in Net::HTTP#connect with socket timeout options


mohamedhafez:
`Timeout.timeout` is inefficient because it spins up a new thread for each invocation, and can be unsafe (see https://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/ and http://blog.headius.com/2008/02/ruby-threadraise-threadkill-timeoutrb.html)

Since we can set open timeouts on sockets now, it would be great to start doing that instead. I sent in a patch years ago, but it was rejected with the reasoning that without `Timeout.timeout`, there would be no way to time out the DNS lookup (https://bugs.ruby-lang.org/issues/12435).

However, it turns out that [Timeout.timeout can't time out the getaddrinfo C call in the first place!](https://bugs.ruby-lang.org/issues/17528) So the status quo is that the `open_timeout` option has no effect on the DNS lookup anyway! Either way we just have to wait for the operating system's `getaddrinfo` call to time out on its own (30 seconds on MacOS 11.1), and setting `open_timeout` has no affect on that at all.

That being the case, it would be great if we could drop the use of `Timeout.timeout` in favor of socket timeout options. I have [a simple patch that is passing all tests](https://github.com/ruby/net-http/compare/master...mohamedhafez:patch-1), and would be happy to send in a pull request, but wanted to discuss a couple things first:

1) To exactly mimic current behavior, we'd have to track the time used up in the DNS lookup, and subtract that from the timeout we give to open the socket. Is this worth it? Is it better done in the implementation of `Socket.tcp`?

2) Should we take advantage of the `:resolv_timeout` available on systems with `getaddrinfo_a` (just Linux I believe?) for Ruby >= 2.7?

deivid-rodriguez:
This was closed by #10, correct?

mohamedhafez:
yes it was. I've actually got equivalent pull requests at https://github.com/ruby/net-smtp/pull/21, https://github.com/ruby/net-pop/pull/3, and https://github.com/ruby/net-ftp/pull/5, @deivid-rodriguez any chance I could get you to give those a quick code review? That seemed to be the impetus behind getting https://github.com/ruby/net-http/pull/10 accepted...

mohamedhafez:
[...]

PR #10, opened 2021-01-16
Replace Timeout.timeout in Net:HTTP#connect

mohamedhafez:
Use socket timeout option instead. 
See https://github.com/ruby/net-http/issues/6 for discussion

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -982,14 +982,13 @@ module Net   #:nodoc:
       end
 
       D "opening connection to #{conn_addr}:#{conn_port}..."
-      s = Timeout.timeout(@open_timeout, Net::OpenTimeout) {
-        begin
-          TCPSocket.open(conn_addr, conn_port, @local_host, @local_port)
-        rescue => e
-          raise e, "Failed to open TCP connection to " +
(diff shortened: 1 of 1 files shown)


mohamedhafez:
Tests are already passing on 2.6 and 2.7, once https://github.com/ruby/net-http/pull/8 gets merged the 3.0 tests pass as well (i tested it locally to be sure).

carlhoerberg:
You need to introduce a `resolv_timeout`, see https://bugs.ruby-lang.org/issues/16381 and https://github.com/ruby/ruby/pull/1806

mohamedhafez:
@carlhoerberg agreed, but I think I'd like to get this PR accepted first and then follow up with another for adding an option for a `:resolv_timeout`. Or feel free to make https://github.com/ruby/ruby/pull/1806 against this repo and I can take this PR down, to be honest until we hear from the maintainers I'm just worried this will be ignored and am hesitant to put any more effort into it.

mohamedhafez:
All the tests are passing now that https://github.com/ruby/net-http/pull/8 got merged:)

deivid-rodriguez (approved):
This changes seems a very good move to me!

PR #16, opened 2021-03-01
Update the content-length heading when decoding bodies

jeremyevans:
Previously, the content-encoding header was removed and the body
was modified, but the content-length header was not modified,
resulting in the content-length header not matching the body
length.

Fixes [Bug #16672]


nurse (approved):
Yeah, when net-http automatically decodes the encoded response, the response should behave as if it's not encoded. The content-length should also be changed.

PR #17, opened 2021-03-09
Add HTTP#force_response_body_encoding for forcing response body encoding

jeremyevans:
This allows for the ability to opt-in to a method to force the
encoding of response bodies.  By setting the accessor to a String
or Encoding instance, it will force the specified encoding.
Setting the value of true will try to detect the encoding of the
response body, either using the Content-Type header (assuming it
specifies charset) or by scanning for a <meta> tag in the document
that specifies the encoding.  The default is false in which case
no forcing of encoding will be done (same as before the patch).

Implements [Feature #2567]
Implements [Feature #15517]

Co-authored-by: Yui Naruse <naruse@ruby-lang.org>

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -690,7 +690,7 @@ module Net   #:nodoc:
       @continue_timeout = nil
       @max_retries = 1
       @debug_output = nil
-      @force_response_body_encoding = false
+      @response_body_encoding = false
 
       @proxy_from_env = false
       @proxy_uri      = nil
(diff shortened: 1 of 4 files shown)


lib/net/http/response.rb
```diff
@@ -245,6 +261,141 @@ def body=(value)
...
+        sniff_encoding(str)
+      end
+    end
+    return encoding
+  end
+
+  # :nodoc:
+  def sniff_encoding(str, encoding=nil)
+    # the encoding sniffing algorithm
+    # http://www.w3.org/TR/html5/parsing.html#determining-the-character-encoding
```
nurse, line 287:
```suggestion
    # https://html.spec.whatwg.org/multipage/parsing.html#encoding-sniffing-algorithm
```

lib/net/http.rb
```diff
@@ -737,6 +738,11 @@ def set_debug_output(output)
     # The local port used to establish the connection.
     attr_accessor :local_port
 
+    # Whether to force a response body encoding.  If true, tries to detect
+    # the response body encoding.  If String or Encoding, uses the specified
+    # encoding.
+    attr_accessor :force_response_body_encoding
+
```
nurse, line 745: The name `force_encoding` is intended to indicate that the method is a dirty, low-level operation. On the other hand, I think this method is highly recommended. The name should reflect the difference.

Also maybe the given String should be verified when it was set instead of the time when it calls `force_encoding`.

nurse:
This also should update `NEWS`.

lib/net/http.rb
jeremyevans: I added commits to fix both of these issues.  I can update NEWS for ruby after this is merged.

nurse (approved):

PR #20, opened 2021-04-28
Add Windows to Actions CI, Ruby 2.5, misc fixes

MSP-Greg:


rhenium:
> Delete Gemfile.lock

Should this commit add it to .gitignore too?

MSP-Greg:
@rhenium

Good idea, probably the reason it was added...  I'll fix

test/net/http/test_https.rb
```diff
@@ -16,14 +16,15 @@ def self.read_fixture(key)
...
 
+  HOST = 'localhost'
   CA_CERT = OpenSSL::X509::Certificate.new(read_fixture("cacert.pem"))
   SERVER_KEY = OpenSSL::PKey.read(read_fixture("server.key"))
   SERVER_CERT = OpenSSL::X509::Certificate.new(read_fixture("server.crt"))
   DHPARAMS = OpenSSL::PKey::DH.new(read_fixture("dhparams.pem"))
   TEST_STORE = OpenSSL::X509::Store.new.tap {|s| s.add_cert(CA_CERT) }
 
   CONFIG = {
-    'host' => '127.0.0.1',
```
rhenium, line 26: Do you have the error output without this change?

I'm wondering if those `rescue SystemCallError` lines scattered over the file are relevant to this. If this (bind to "localhost") works, can they be removed safely now?

MSP-Greg: They may be.  Current mingw jobs in ruby/ruby are failing because of this (see https://github.com/ruby/ruby/actions).  I don't have any CI logs in my fork.  I fixed locally, then pushed...

I didn't look at any blame/history, but if it's never been tested on Windows...

rhenium: From https://github.com/ruby/ruby/runs/2456377434?check_suite_focus=true:

```
   1) Failure:
TestNetHTTPS#test_session_reuse [D:/a/ruby/ruby/src/test/net/http/utils.rb:48]:
<[]> expected but was
<["[2021-04-28 10:35:33] ERROR Errno::ECONNABORTED: An established connection was aborted by the software in your host machine.\n" +
 "\tD:/a/ruby/ruby/build/.ext/common/openssl/buffering.rb:80:in `sysread'\n" +
 "\tD:/a/ruby/ruby/build/.ext/common/openssl/buffering.rb:80:in `fill_rbuff'\n" +
 "\tD:/a/ruby/ruby/build/.ext/common/openssl/buffering.rb:323:in `eof?'\n" +
 "\tD:/a/ruby/ruby/src/tool/lib/webrick/httpserver.rb:82:in `run'\n" +
 "\tD:/a/ruby/ruby/src/tool/lib/webrick/server.rb:310:in `block in start_thread'\n"]>.

  2) Failure:
TestNetHTTPS#test_session_reuse [D:/a/ruby/ruby/src/test/net/http/test_https.rb:166]:
<true> expected but was
<false>.
```

Thanks for the pointer, but this puzzles me... I was expecting the failing test cases to be `test_get_SNI` and alike, which were only recently added and didn't have `rescue SystemCallError` yet. (FWIW, the explanation for `rescue` is in https://github.com/ruby/net-http/commit/25048476d026ea64bb3dc3607ba7b48f8836a5d1.)

MSP-Greg: I hate this.  I pushed the file change to ruby-loco and my ruby/ruby fork.  test-all passed everywhere.

But, in test-spec, it timing out in ruby-loco, and spec/ruby/library/net/http/http/active_spec.rb is failing, maybe more...

MSP-Greg: Re the failures, thanks for looking further.  I can revert that commit and the other `rescue SystemCallError` code?

I'm looking at the spec failures.

MSP-Greg: @rhenium

I looked at and removed all the `rescue SystemCallError` code, passed in my fork.  Re core test-spec, it appears to be the same issue, mixing `localhost` & `127.0.0.1`.  Running in my fork right now, I probably do a two commit PR to ruby/ruby if it's ok.  Thanks for your help.  JFYI, had similar issue in Puma.

rhenium: `TestNetHTTPS#test_session_reuse` failure is reliably reproduced on my local computer by inserting `sleep 2.5` (> `keep_alive_timeout`, which is 2 seconds by default) right before the second `http.get("/")`. There actually seems to be an issue with session resumption in net/http. Looking into this...

```diff
diff --git a/test/net/http/test_https.rb b/test/net/http/test_https.rb
index e9aee15bd31f..21bba4689643 100644
--- a/test/net/http/test_https.rb
+++ b/test/net/http/test_https.rb
@@ -160,6 +160,7 @@ def test_session_reuse
     http.finish
 
     http.start
+    sleep 2.5
     http.get("/")
 
     socket = http.instance_variable_get(:@socket).io
```

rhenium: Regarding localhost vs 127.0.0.1 and rescue, I /think/ the changes in this Pull Request are correct, but I can't confirm - I don't know in which environment it is actually needed. Without the changes, `TestNetHTTPS#test_min_version` must fail in such setup, but we have no such a setup in our CI (rubyci.org / GitHub Actions).

<details>
`test_get_SNI` was a wrong example. It does `http.ipaddr = config('host')`, so it was fine as is.
</details>

ruby/ruby's Github Actions failure on mingw (`TestNetHTTPS#test_session_reuse`) looks like a separate issue (as I wrote in my last comment).

MSP-Greg: @rhenium

Thanks for your additional investigation.  I saw ruby-loco failed, then went to ruby/ruby, and saw Ubuntu & macOS passing (along with CI here), but failing mingw, I assumed it was probably a test issue.  Didn't even look at lib code here.  I did review the code in ruby/openssl test_ssl_session.rb.

Anyway, re the `sleep 2.5` addition, I think `http.finish` is closing the socket, and the handshake occurs when `http.start` is called (it calls `connect_nonblock`)?  So I'm not sure what the sleep is affecting, as I would think `keep_alive_timeout` would be different than a 'data' timeout, but I haven't looked...

Also, wondering whether to define 'HOST_IP = '127.0.0.1' and use it for all refs in the test?

rhenium: https://github.com/ruby/net-http/pull/21 should fix `TestNetHTTPS#test_session_reuse`, though running the test suite takes 10 minutes on Windows. This is due to the slow fallback from IPv6 (localhost) to IPv4 (127.0.0.1) and this (your) Pull Request should resolve it.

Session resumption was not done correctly because the test case was establishing 3 TLS connections in total instead of the intended 2 due to the bug & a change in OpenSSL 1.1.1 which prohibited using a session ticket more than once in TLS 1.3 clients (https://github.com/openssl/openssl/pull/6601).

rhenium (approved):

test/net/http/test_https.rb
MSP-Greg: > change in OpenSSL 1.1.1

Thanks.  I vaguely recalled something about that, but my finding it might have taken a while...

Should I change `test_session_reuse` as above?  `session_reused?` is important, but the real test is can the socket be used, and hence, the body assert?

MSP-Greg: @rhenium

I removed a Rakefile change that wasn't needed.  I revised the test as above, and also added the constant.  Changed it to an IPv6 address locally, and all passed.

And, since I haven't contributed here previously, it needs approval for CI.  Passed in my fork.  Thanks.

MSP-Greg:
Since this changes a few files that are also changed in #19, I added Ruby 2.5 to Actions, and listed @deivid-rodriguez as a co-author.

Testing was being run with `rake test`, changed to `bundle exec rake test` since `bundle install` is used.

deivid-rodriguez:
I'm still in favour of releasing a gemified version of net-http still supporting ruby 2.5, and then dropping support, but it sounds unrelated to this PR.

MSP-Greg:
This PR does not change any 'lib' files, only files related to testing and building.  Using Ruby 2.5 is part of that.

I'll be happy to revert those changes.

This PR fixes an issue that has happened before, and that is a ruby repo (std-lib, default or bundled gem) breaking CI in ruby/ruby, which has been failing since 27-Apr...

deivid-rodriguez:
Restoring support for ruby 2.5 seems unrelated to getting the CI back to green. The fact that two separate PRs change the same part of a repository doesn't mean they are related or should be introduced together.

That said, I don't make any decisions here, but again, I'd be very happy if ruby 2.5 support is restored. I just feel it unnecessarily makes this PR more controversial :man_shrugging:.

We'll say what maintainers say.

hsbt:
Fixed at https://github.com/ruby/net-http/pull/181

Replacing `localhost` to `HOST` and `127.0.0.1` to `HOST_IP` is seems good. If you still interest that, can you submit another PR? Thanks.

PR #24, opened 2021-05-01
Net::HTTP.get_response can receive URI as string

kyanagi:
We need to pass a `URI` object to `Net::HTTP.get_response` (or `Net::HTTP.get`).

```ruby
Net::HTTP.get_response(URI('http://www.example.com/index.html'))
```

With this patch, these methods can receive a URI-string.

```ruby
Net::HTTP.get_response('http://www.example.com/index.html')
```


k0kubun:
For consistency, supporting the same thing for `Net::HTTP.post` as well might be useful.

kyanagi:
@k0kubun
Thank you for your suggestion.
I added the same support for `Net::HTTP.post` and `Net::HTTP.post_form`.

tachyons:
This looks cool, @k0kubun Does this have any blockers ?

k0kubun:
I left a comment because I contributed a similar feature, but I'm not the maintainer. You need to ping the actual maintainer https://github.com/ruby/ruby/blob/master/doc/maintainers.md#libnethttprb-libnethttpsrb after fixing the conflict.

PR #25, opened 2021-05-09
Add path and headers to Request inspect method

adam12:
I find having the path and headers when inspecting a request to be handy.

```
>> Net::HTTP::Get.new "/"
=> #<Net::HTTP::Get GET path="/" headers={"accept-encoding"=>["gzip;q=1.0,deflate;q=0.6,identity;q=0.3"], "accept"=>["*/*"], "user-agent"=>["Ruby"]}>
```


nobu:
How about to add `pretty_print` and keeping `inspect` as-is, like #160?

adam12:
@nobu seems reasonable to me.

adam12:
Closing in favour of nobu's suggestion.

PR #27, opened 2021-06-11
Enforce write timeout when body_stream is used

miguelfteixeira:
The existing implementation of `Net::HTTP#write_timeout` relies on
`Net::BefferedIO` to trigger the `Net::WriteTimeout` error. This commit
changes `send_request_with_body_stream` to remove the optimization that
was making `Net::HTTP#write_timeout` not work when `body_stream` is
used.

This issue affects multipart requests from third-party libraries that wrap
`Net::HTTP` like `Faraday`.

Open issue:
https://bugs.ruby-lang.org/issues/17933

### DIFF
--- a/lib/net/http/generic_request.rb
+++ b/lib/net/http/generic_request.rb
@@ -202,9 +202,7 @@ class Net::HTTPGenericRequest
       IO.copy_stream(f, chunker)
       chunker.finish
     else
-      # copy_stream can sendfile() to sock.io unless we use SSL.
-      # If sock.io is an SSLSocket, copy_stream will hit SSL_write()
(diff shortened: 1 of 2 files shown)


jeremyevans (approved):
This looks good to me, and I think it is worth dropping the optimization in order to fix the bug.  However, this is ultimately up to @nurse .  I think even if we don't disable the optimization by default, we should at least offer the option to disable it.

nurse:
Well, I worried that this disables the optimization of IO.copy_stream, but [byroot's comment](https://bugs.ruby-lang.org/issues/17933#note-1) "this optimization is already disabled for SSL sockets" convinced me. Go ahead!

PR #30, opened 2021-09-03
Update Gemfile.lock so tests can run out of the box

jarl-dk:


hsbt:
@jarl-dk Thanks, But lockfile existence is my mistake. I've removed it at https://github.com/ruby/net-http/commit/b81bd7955d978b365020e29a4cbb8a5f047cd8d1

PR #36, opened 2022-01-04
Do not set SNI hostname if connecting to IP address

jeremyevans:
RFC 6066, section 3, explicitly disallows the use of an IP address
as an SNI server name.  So check if the connection is being made
to an IP address using the resolv regexps, and do not set an SNI
hostname in that case.

Note that this results in a warning in the tests, but that is due
to an issue in the openssl extension.  A pull request has been
submitted to the openssl extension to remove that warning.

Revert the previous change that modified the regexp used for
checking the error message.

This not only fixes net/http tests, it also fixes tests for
webrick and open-uri when used with recent LibreSSL , which both make SSL connections to 127.0.0.1
in their tests.

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -22,6 +22,7 @@
 
 require 'net/protocol'
 require 'uri'
+require 'resolv'
 autoload :OpenSSL, 'openssl'
 
 module Net   #:nodoc:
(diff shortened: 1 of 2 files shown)


jeremyevans:
I've pushed a new version of this that doesn't cause any warnings, by unsetting `verify_hostname` if set and connecting to an IP address, but still keeping the `post_connection_check` to make sure the certificate provided matches the IP address.  I plan to merge this in one week if there are no objections.

PR #39, opened 2022-01-16
Rename `D` to `debug` in `Net::HTTP`

baweaver:
Renames `D` to `debug` in `Net::HTTP` and introduces an alias for
backwards compatibility. This was done for readability reasons, in that
`D` did not clearly reflect what the method was doing and can cause some
confusion.

Originally in Ruby repo, moved PR here from old one: https://github.com/ruby/ruby/pull/3714

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -993,7 +993,7 @@ module Net   #:nodoc:
         conn_port = port
       end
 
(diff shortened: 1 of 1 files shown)


hsbt:
@baweaver Thanks!

PR #42, opened 2022-01-28
parsing add sensitive headers with '@'

GodKelvin:
Because in some applications the header is sensitive,  for this, if I pass an '@', I make the same value in header, without capitalize. For example, I have a request for API make in C#, him break if I pass 'Token' but worked if I pass 'token', because he not follow the RFC.

### DIFF
--- a/lib/net/http/header.rb
+++ b/lib/net/http/header.rb
@@ -218,7 +218,15 @@ module Net::HTTPHeader
   alias canonical_each each_capitalized
 
   def capitalize(name)
-    name.to_s.split(/-/).map {|s| s.capitalize }.join('-')
+    #Because in some applications the header is sensitive, 
+    #for this, if I pass an '@', I make the same value in header, without capitalize.
+    #For example, I have a request for API make in C#, him break if I pass 'Token' but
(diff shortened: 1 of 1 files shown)


lib/net/http/header.rb
```diff
@@ -218,7 +218,15 @@ def each_capitalized
   alias canonical_each each_capitalized
 
   def capitalize(name)
-    name.to_s.split(/-/).map {|s| s.capitalize }.join('-')
+    #Because in some applications the header is sensitive, 
+    #for this, if I pass an '@', I make the same value in header, without capitalize.
+    #For example, I have a request for API make in C#, him break if I pass 'Token' but
+    #worked if I pass 'token', because he not follow the RFC.
+    if(name[0] == '@')
+      name.gsub('@', '')
```
nobu, line 226: Looks an interesting idea, but ambiguous a little.
This checks the first letter only but deletes all `@`.
What do you expect when `name` contains an `@` in the middle?

kahshing96:
hi @GodKelvin @nobu is there any update on this PR? would like to have this feature be merged which will be helpful in certain case where the 3rd-party API server is not following the RFC standard

PR #45, opened 2022-03-23
Ignore EBADF when closing a socket after exception

mohamedhafez:
On JRuby, Errno::EBADF is thrown sometimes when trying to close the socket after an exception has occurred. This change ignores that so that the original exception can be raised

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -1078,7 +1078,10 @@ module Net   #:nodoc:
     rescue => exception
       if s
         debug "Conn close because of connect error #{exception}"
-        s.close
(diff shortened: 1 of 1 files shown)


mohamedhafez:
This is still an issue even with the latest release, and I still have to monkeypatch on JRuby to get around this issue. Just a little ping if someone could please take a look at this

headius:
I'm reviewing CRuby code to see why they don't raise this error.

mohamedhafez:
The issue has been fixed in JRuby, and this PR is no longer necessary, thanks headius!

PR #52, opened 2022-04-08
feature detect to make net/http usable with JRuby

kares:
up till JRuby 9.3 (Ruby 2.6 compatibility) JRuby shipped a slightly patched version of net/http to work-around some of the limitations - mostly of JRuby-OpenSSL which is implemented on top of Java APIs (does not use OpenSSL).

for JRuby 9.4, which is in development atm, it would be nice to be able to ship the gem version of net/http - the changes here represent an effort to make things compatible, there's really one important change:

- `SSLContext#session_new_cb` in not implemented under JRuby, it could be a dummy no-op but this would confuse users as these session_xxx callbacks do not get actually called (no way to hook into session management on the Java SSL engine end)
- `SSLContext#session_cache_mode=` historically exist in JRuby but is effectively a no-op and issues a `Kernel.warn`, the `SSLContext#session_cache_mode` reader always returns `nil` as an attempt to be able to "feature-detect"
- also looked into low hanging fruit in some of the tests (with these there's only 3 failures left under JRuby)

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -1033,9 +1033,11 @@ module Net   #:nodoc:
           end
         end
         @ssl_context.set_params(ssl_parameters)
-        @ssl_context.session_cache_mode =
-          OpenSSL::SSL::SSLContext::SESSION_CACHE_CLIENT |
-          OpenSSL::SSL::SSLContext::SESSION_CACHE_NO_INTERNAL_STORE
+        unless @ssl_context.session_cache_mode.nil? # a dummy method on JRuby
+          @ssl_context.session_cache_mode =
+              OpenSSL::SSL::SSLContext::SESSION_CACHE_CLIENT |
+                  OpenSSL::SSL::SSLContext::SESSION_CACHE_NO_INTERNAL_STORE
+        end
         if @ssl_context.respond_to?(:session_new_cb) # not implemented under JRuby
           @ssl_context.session_new_cb = proc {|sock, sess| @ssl_session = sess }
         end
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -1036,7 +1036,9 @@ module Net   #:nodoc:
         @ssl_context.session_cache_mode =
           OpenSSL::SSL::SSLContext::SESSION_CACHE_CLIENT |
           OpenSSL::SSL::SSLContext::SESSION_CACHE_NO_INTERNAL_STORE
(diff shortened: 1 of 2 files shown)


lib/net/http.rb
```diff
@@ -1033,10 +1033,14 @@ def connect
           end
         end
         @ssl_context.set_params(ssl_parameters)
-        @ssl_context.session_cache_mode =
-          OpenSSL::SSL::SSLContext::SESSION_CACHE_CLIENT |
-          OpenSSL::SSL::SSLContext::SESSION_CACHE_NO_INTERNAL_STORE
-        @ssl_context.session_new_cb = proc {|sock, sess| @ssl_session = sess }
+        unless @ssl_context.session_cache_mode.nil? # a dummy method on JRuby
```
jeremyevans, line 1036: If JRuby doesn't support `@ssl_context.session_cache_mode`, maybe don't define the method and use `if defined?(@ssl_context.session_cache_mode)`.

jeremyevans (commented):
This looks OK to me, though I don't have a good knowledge of JRuby's openssl extension.

lib/net/http.rb
headius: Yes, we probably should do that, but we implemented it as a no-op plus warning so we might be reluctant to remove the method all together.

kares: :+1: the reasoning being `session_cache_mode` does not affect the user-code much overall and thus it existed for some time and removing it might be considered breaking ... thus the `nil` compromise will probably stay with us.
smt like `session_new_cb`, while being "low" level, tends to get much more critical in user-code - if we had it as a dummy attr and the callback never fired it would be counter productive ...

headius:
I think this is ready to merge. It would be helpful to us completing work on JRuby 9.4 if we could get this into a gem (possibly prerelease).

nurse (approved):
Generally it looks good to me.
@jeremyevans, could you merge this if you also think loos good

headius:
Can we get a gem pushed (preview is ok) so we can hook it up to JRuby's build and start using the gem?

headius:
Ping @nurse @hsbt can we get a gem release or preview gem release with this change?

hsbt:
@headius I send an invitation of net-http owner. Can you confirm it?

headius:
@hsbt thank you! I will push a pre gem today.

PR #55, opened 2022-05-09
implement talking SSL to the proxy too

evgeni:
https://bugs.ruby-lang.org/issues/16482

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -1103,7 +1103,7 @@ module Net   #:nodoc:
     # For proxy-defining arguments +p_addr+ through +p_no_proxy+,
     # see {Proxy Server}[rdoc-ref:Net::HTTP@Proxy+Server].
     #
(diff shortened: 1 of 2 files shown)


lib/net/http.rb
```diff
@@ -1024,7 +1028,13 @@ def connect
       debug "opened"
       if use_ssl?
         if proxy?
-          plain_sock = BufferedIO.new(s, read_timeout: @read_timeout,
+          if @proxy_use_ssl
+            proxy_sock = OpenSSL::SSL::SSLSocket.new(s)
```
evgeni, line 1032: this probably needs some more work, especially when it comes to verification etc, but it's a start :)

sbernhard:
Would be good if someone, like @peterzhu2118 or @BurdetteLamar can leave a commit for this change?

lib/net/http.rb
BurdetteLamar: @evgeni, I try to avoid code changes.

evgeni: Not sure what you mean by that?

hsbt:
👍

evgeni:
Thanks!

PR #58, opened 2022-05-17
SSL - Add alpn_protocols to SSLContext attributes

MSP-Greg:
Since the code used with `SSL_IVNAMES` and `SSL_ATTRIBUTES` is tested, there isn't a need for a specific test, as it's setting SSLContext attributes.


MSP-Greg:
Update comment for `HTTP.start`

PR #59, opened 2022-05-18

BurdetteLamar:
Treats:
- ::get
- ::get_print
- ::get_response

Fixes a broken link in the class doc; clarifies example usage.

I'm proposing to use example site [JSONPlaceholder](https://jsonplaceholder.typicode.com) because it can return responses short enough for us to show, unlike site [Example Domain](http://www.example.com), which always returns something large.

If this is approved, there will be much more.


k0kubun:
I'm okay with this change, but I'm probably not authorized to give approval since I'm not a maintainer here.

PR #65, opened 2022-08-20
Enable frozen_string_literal

casperisfine:
Net::HTTP contain quite a few string literals that would benefit from being interned.

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -1,4 +1,4 @@
-# frozen_string_literal: false
+# frozen_string_literal: true
 #
 # = net/http.rb
 #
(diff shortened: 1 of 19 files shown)


jeremyevans (commented):
The implementation looks fine to me, but this change probably needs to be approved by @nurse.

baweaver:
Have we taken a look at merging this one? Probably needs to be rebased though.

technicalpickles:
@baweaver and I were pairing on some performance issues last month with memory_profiler, and stumbled upon net/http in a few spots

For our one system spec, we're seeing `"-"` allocated 11374 times, and `""` 4563 times in `net/http/header`. Plus 3425 allocations of `" "`and 5701 allocations of `""` in `net/http/response`. Using frozen string literally would take care of all those.

hsbt:
`frozen_string_literal` is enabled by https://github.com/ruby/net-http/pull/144

Issue #68, opened 2022-09-06
fail to fetch proxy user/pw from env variable


kaisecheng:
[http.rb](https://github.com/ruby/net-http/blob/v0.2.2/lib/net/http.rb#L1225) use RUBY_PLATFORM to decide the source of username/ password of the proxy. RUBY_PLATFORM of jruby is `java`. It blocks jruby user from using proxy with env variable

ruby --version 
jruby 9.3.6.0 (2.6.8) 2022-06-27 7a2cbcd376 OpenJDK 64-Bit Server VM 11.0.15+10 on 11.0.15+10 +jit [x86_64-darwin]

### reproduce steps
1. export JRUBY_OPTS='-J-Dhttps.proxyHost=user:pw@host -J-Dhttps.proxyPort=3128 -J-Dhttp.proxyHost=user:pw@host -J-Dhttp.proxyPort=3128'
2. ruby -e "require 'net/http'; uri = URI('https://www.google.com'); Net::HTTP.get(uri);"

### log
```
Net::HTTPServerException: 407 "Proxy Authentication Required"
        error! at /Users/kcheng/.rbenv/versions/jruby-9.3.6.0/lib/ruby/stdlib/net/http/response.rb:122
         value at /Users/kcheng/.rbenv/versions/jruby-9.3.6.0/lib/ruby/stdlib/net/http/response.rb:131
       connect at /Users/kcheng/.rbenv/versions/jruby-9.3.6.0/lib/ruby/stdlib/net/http.rb:973
      do_start at /Users/kcheng/.rbenv/versions/jruby-9.3.6.0/lib/ruby/stdlib/net/http.rb:934
         start at /Users/kcheng/.rbenv/versions/jruby-9.3.6.0/lib/ruby/stdlib/net/http.rb:923
         start at /Users/kcheng/.rbenv/versions/jruby-9.3.6.0/lib/ruby/stdlib/net/http.rb:605
  get_response at /Users/kcheng/.rbenv/versions/jruby-9.3.6.0/lib/ruby/stdlib/net/http.rb:481
           get at /Users/kcheng/.rbenv/versions/jruby-9.3.6.0/lib/ruby/stdlib/net/http.rb:458
        <main> at -e:1
```

PR #69, opened 2022-09-13
add java as a valid RUBY_PLATFORM 

kaisecheng:
jruby users are not able to use env variable to setup proxy with basic authentication because the RUBY_PLATFORM does not allow `java`.
The commit includes `java` as a valid platform.

### expected behaviour 
1. export JRUBY_OPTS='-J-Dhttps.proxyHost=user:pw@host -J-Dhttps.proxyPort=3128 -J-Dhttp.proxyHost=user:pw@host -J-Dhttp.proxyPort=3128'
2. ruby -e "require 'net/http'; uri = URI('https://www.google.com'); Net::HTTP.get(uri);"
3. no error is thrown

Fixed: #68

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -1222,7 +1222,7 @@ module Net   #:nodoc:
     end
 
     # [Bug #12921]
-    if /linux|freebsd|darwin/ =~ RUBY_PLATFORM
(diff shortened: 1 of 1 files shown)


hsbt:
@kaisecheng Thanks, But this condition has been removed at https://github.com/ruby/net-http/pull/66

Issue #70, opened 2022-09-22
Net::HTTPResponse has bad nil handling in read_body and stream_check


BrianHawley:
I was getting a weird error from Net::HTTPResponse, and noticed that a couple methods had some bad nil checking.

In `read_body`, it sets `@body = nil` and then later on calls string methods on it like `@body.force_encoding`. Right after you set `@read = true` you should add this code: `return if @body.nil?`.

Then, the weird error was from `stream_check`:
```
undefined method `closed?' for nil:NilClass (NoMethodError)

    raise IOError, 'attempt to read body out of block' if @socket.closed?
```
Since you set `@socket = nil` in `reading_body`, the `stream_check` method should be:
```ruby
def stream_check
  raise IOError, 'attempt to read body out of block' if @socket.nil? || @socket.closed?
end
```

BrianHawley:
Here it is in patch form, for those who prefer that:
```ruby
# frozen_string_literal: true

require "net/http"

begin
  response = Net::HTTPResponse.new("", "", "")
  response.body
rescue IOError
  # We're good
rescue
  # Bad exception
  Net::HTTPResponse.class_exec do
    private

    def stream_check
      raise IOError, "attempt to read body out of block" if @socket.nil? || @socket.closed?
    end
  end
end
```

PR #71, opened 2022-09-22
Net::HTTPResponse nil checking

BrianHawley:
Fix nil handling in read_body and stream_check.

Fixes: #70

### DIFF
--- a/lib/net/http/response.rb
+++ b/lib/net/http/response.rb
@@ -231,6 +231,7 @@ class Net::HTTPResponse
       @body = nil
     end
     @read = true
+    return if @body.nil?
 
     case enc = @body_encoding
     when Encoding, false, nil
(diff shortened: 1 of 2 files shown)


BrianHawley:
@nobu or @jeremyevans what do you think? The tests pass locally. Just improving error handling.

jeremyevans:
The `stream_check` change looks fine. It's questionable whether `body` should return `nil` or an empty string in such an error case.  That's probably a decision that @nurse should make.

BrianHawley:
@jeremyevans if it should return an empty string, then we should assign the empty string a few lines above instead of assigning nil. At that point, the line I added can be removed and the test can be adjusted accordingly.

BrianHawley:
@jeremyevans and @nurse what do I need to do to get this moving forward? Should I split the `read_body` and `stream_check` changes into separate PRs? I made this PR in September, so I'm interested in getting this in.

jeremyevans:
@BrianHawley  I don't think there is more you can do. @hsbt @nobu Can either of you review/merge this?

BrianHawley:
Can this get merged in time to get it into Ruby 3.3? I made sure to make the PR in time for it to get into Ruby 3.2.

PR #77, opened 2022-11-15

BurdetteLamar:
- Revises most of class doc for Net::HTTP.  (Does not modify sections "Following Redirection" and after.)
- Treats a few methods, including Net::HTTP.start.
- Adds to class docs for Net::HTTPRequest and Net::HTTPResponse.

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -32,110 +32,290 @@ module Net   #:nodoc:
   class HTTPHeaderSyntaxError < StandardError; end
   # :startdoc:
 
-  # == An HTTP client API for Ruby.
+  # \Class \Net::HTTP provides a rich library that implements the client
+  # in a client-server model that uses the \HTTP request-response protocol.
+  # For information about \HTTP, see
   #
-  # Net::HTTP provides a rich library which can be used to build HTTP
-  # user-agents.  For more details about HTTP see
-  # [RFC2616](http://www.ietf.org/rfc/rfc2616.txt).
+  # - {Hypertext Transfer Protocol}[https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol].
+  # - {Technical overview}[https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Technical_overview].
   #
-  # Net::HTTP is designed to work closely with URI.  URI::HTTP#host,
-  # URI::HTTP#port and URI::HTTP#request_uri are designed to work with
-  # Net::HTTP.
+  # Note: If you are performing only a few GET requests, consider using
+  # {OpenURI}[https://docs.ruby-lang.org/en/master/OpenURI.html];
+  # otherwise, read on.
   #
-  # If you are only performing a few GET requests you should try OpenURI.
+  # == Synopsis
   #
-  # == Simple Examples
+  # If you are already familiar with \HTTP, this synopsis may be helpful.
   #
-  # All examples assume you have loaded Net::HTTP with:
+  # {Session}[rdoc-ref:Net::HTTP@Sessions] with multiple requests for
+  # {HTTP methods}[https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Request_methods]:
+  #
+  #   Net::HTTP.start(hostname) do |http|
+  #     # Session started automatically before block execution.
+  #     http.get(path_or_uri, headers = {})
+  #     http.head(path_or_uri, headers = {})
+  #     http.post(path_or_uri, data, headers = {})  # Can also have a block.
+  #     http.put(path_or_uri, data, headers = {})
+  #     http.delete(path_or_uri, headers = {Depth: 'Infinity'})
+  #     http.options(path_or_uri, headers = {})
+  #     http.trace(path_or_uri, headers = {})
+  #     http.patch(path_or_uri, data, headers = {}) # Can also have a block.
+  #     # Session finished automatically at block exit.
+  #   end
+  #
+  # {Session}[rdoc-ref:Net::HTTP@Sessions] with multiple requests for
+  # {WebDAV methods}[https://en.wikipedia.org/wiki/WebDAV#Implementation]:
+  #
+  #   Net::HTTP.start(hostname) do |http|
+  #     # Session started automatically before block execution.
+  #     http.copy(path_or_uri, headers = {})
+  #     http.lock(path_or_uri, body, headers = {})
+  #     http.mkcol(path_or_uri, body = nil, headers = {})
+  #     http.move(path_or_uri, headers = {})
+  #     http.propfind(path_or_uri, body = nil, headers = {'Depth' => '0'})
+  #     http.proppatch(path_or_uri, body, headers = {})
+  #     http.unlock(path_or_uri, body, headers = {})
+  #     # Session finished automatically at block exit.
+  #   end
+  #
+  # Each of the following methods automatically starts and finishes
+  # a {session}[rdoc-ref:Net::HTTP@Sessions] that sends a single request:
+  #
+  #   # Return string response body.
+  #   Net::HTTP.get(hostname, path, port = 80)
+  #   Net::HTTP.get(uri, headers = {}, port = 80)
+  #
+  #   # Write string response body to $stdout.
+  #   Net::HTTP.get_print(hostname, path_or_uri, port = 80)
+  #   Net::HTTP.get_print(uri, headers = {}, port = 80)
+  #
+  #   # Return response as Net::HTTPResponse object.
+  #   Net::HTTP.get_response(hostname, path_or_uri, port = 80)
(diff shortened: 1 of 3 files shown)


lib/net/http/response.rb
```diff
@@ -1,20 +1,156 @@
...
+#
+# Examples here assume that <tt>net/http</tt> has been required
+# (which also requires +uri+):
+#
+#   require 'net/http'
+#
+# Many code examples here use these example websites:
+#
+# - https://jsonplaceholder.typicode.com.
+# - http:example.com.
```
peterzhu2118, line 15:
```suggestion
# - http://example.com.
```

lib/net/http.rb
```diff
@@ -32,110 +32,290 @@ class HTTPBadResponse < StandardError; end
...
   #   require 'net/http'
   #
-  # This will also require 'uri' so you don't need to require it separately.
+  # Many code examples here use these example websites:
   #
-  # The Net::HTTP methods in the following section do not persist
-  # connections.  They are not recommended if you are performing many HTTP
-  # requests.
+  # - https://jsonplaceholder.typicode.com.
+  # - http:example.com.
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed (3 places).  Should "About the Examples" be pushed out to an includable file (in a future PR)?  It's now used in three places, and may be in more.

lib/net/http.rb
```diff
@@ -32,110 +32,290 @@ class HTTPBadResponse < StandardError; end
...
+  # On the internet, a URI
+  # ({Universal Resource Identifier}[https://en.wikipedia.org/wiki/Uniform_Resource_Identifier])
+  # is a string that identifies a particular resource.
+  # It consists of some or all of: scheme, hostname, path, query, and fragment;
+  # see {URI syntax}[https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax].
   #
-  #   uri = URI('http://www.example.com/search.cgi')
-  #   res = Net::HTTP.post_form(uri, 'q' => 'ruby', 'max' => '50')
-  #   puts res.body
+  # A Ruby {URI::Generic}[]https://docs.ruby-lang.org/en/master/URI/Generic.html] object
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed.

lib/net/http.rb
```diff
@@ -32,110 +32,290 @@ class HTTPBadResponse < StandardError; end
...
+  #   File.write(path, '') unless File.exist?(path)
+  #   file = File.stat(path)
+  #   req = Net::HTTP::Get.new(uri)
+  #   req['If-Modified-Since'] = file.mtime.rfc2822
+  #   res = Net::HTTP.start(hostname) do |http|
+  #     http.request(req)
+  #   end
+  #   if res.is_a?(Net::HTTPSuccess)
+  #     File.write(path, res.body)
+  #   end
```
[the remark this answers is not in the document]
BurdetteLamar: This is adapted from the only such example in the previous doc.  I don't think it's needed at all.  There are dozens of header fields whose meanings we don't explicate here.

Btw, example.com is even dumber than jsonplaceholder.typicode.com, which is why I'm changing to the latter.

lib/net/http.rb
```diff
@@ -441,25 +463,12 @@ class << HTTP
       alias is_version_1_2? version_1_2?   #:nodoc:
     end
 
+    # :call-seq:
+    #   Net::HTTP.get_print(hostname, path, port = 80) -> nil
+    #   Net::HTTP:get_print(uri, headers = {}, port = 80) -> nil
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed (3 places).

lib/net/http.rb
```diff
@@ -469,40 +478,54 @@ def HTTP.get_print(uri_or_host, path_or_headers = nil, port = nil)
...
-    end
-
-    # Sends a GET request to the target and returns the HTTP response
-    # as a Net::HTTPResponse object.  The target can either be specified as
-    # (+uri+, +headers+), or as (+host+, +path+, +port+ = 80); so:
+    #   {
+    #     "userId": 1,
+    #     "id": 1,
+    #     "title": "delectus aut autem",
+    #     "completed":
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed.

lib/net/http.rb
```diff
@@ -469,40 +478,54 @@ def HTTP.get_print(uri_or_host, path_or_headers = nil, port = nil)
...
+    #   headers = {Accept: 'text/html'}
+    #   puts Net::HTTP.get(uri, headers)
     #
-    #    res = Net::HTTP.get_response('www.example.com', '/index.html')
-    #    print res.body
+    # Output:
     #
-    # you can also specify request headers:
+    #   {
+    #    "userId": 1,
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed.

lib/net/http.rb
```diff
@@ -588,36 +611,78 @@ def HTTP.socket_type   #:nodoc: obsolete
       BufferedIO
     end
 
-    # :call-seq:
-    #   HTTP.start(address, port, p_addr, p_port, p_user, p_pass, &block)
-    #   HTTP.start(address, port=nil, p_addr=:ENV, p_port=nil, p_user=nil, p_pass=nil, opt, &block)
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed.

lib/net/http/response.rb
BurdetteLamar: Fixed.

BurdetteLamar:
Wrt headers, they have no apparent effect on either host we've mentioned.  Would it be useful to introduce another "real" host (github rest api, for example) to have more "real" examples? (But not in this PR.)

lib/net/http/request.rb
```diff
@@ -1,8 +1,82 @@
...
+# === \Net::HTTP::Post
+#
+# A POST request may be sent using request class \Net::HTTP::Post:
+#
+#   json = {title: 'foo', body: 'bar', userId: 1}
+#   # => "{\"title\":\"foo\",\"body\":\"bar\",\"userId\":1}"
+#   _uri = uri.dup
+#   _uri.path = '/posts'
+#   req = Net::HTTP::Post.new(_uri) # => #<Net::HTTP::Post POST>
+#   req.body = json
```
peterzhu2118, line 62: `req.body` should be a string rather than a hash I think.

```suggestion
#   _uri = uri.dup
#   _uri.path = '/posts'
#   req = Net::HTTP::Post.new(_uri) # => #<Net::HTTP::Post POST>
#   req.body = '{"title": "foo", "body": "bar", "userId": 1}'
```

peterzhu2118 (commented):
Maybe something like https://httpbin.org/ that is more of a "swiss army knife" of endpoints to test various things would be useful (to demo things like headers, authentication, status codes, redirect, etc.).

BurdetteLamar:
> Maybe something like https://httpbin.org/ that is more of a "swiss army knife" of endpoints to test various things would be useful (to demo things like headers, authentication, status codes, redirect, etc.).

I'll look into that.  (But not for this PR?)

peterzhu2118 (approved):

BurdetteLamar:
A reader pointed out (offline) that I'd added to request.rb a new section for each of the 15 subclasses, but with text and examples for only three.  For some (perhaps most) of the requests, the example host will not behave well; so I have removed the new sections altogether.

lib/net/http/request.rb
```diff
@@ -1,8 +1,26 @@
...
+# - \Net::HTTP::Trace
+# - \Net::HTTP::Patch
+# - \Net::HTTP::Put
+# - \Net::HTTP::Copy
+# - \Net::HTTP::Lock
+# - \Net::HTTP::Mkcol
+# - \Net::HTTP::Move
+# - \Net::HTTP::Propfind
+# - \Net::HTTP::Proppatch
+# - \Net::HTTP::Unlock
```
peterzhu2118, line 23: Should we remove the escaping so that it links to the subclasses?

peterzhu2118, line 23: How is this list sorted? Should we sort it alphabetically (or some other way) so it's more intuitive to read?

BurdetteLamar: The subclasses' docs are empty(ish).

BurdetteLamar: They're in the order as given in the Wikipedia docs we've linked (which I assumed is the same as in the underlying specs):
- https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Request_methods
- https://en.wikipedia.org/wiki/WebDAV#Implementation
Want them alphabetical?

peterzhu2118: I see, that makes sense. I wonder how Wikipedia orders their list 🤔

peterzhu2118 (approved):

PR #78, opened 2022-11-18
[DOC] About the Examples moved to separate file

BurdetteLamar:
Included elsewhere twice; there will be more.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 /.yardoc
 /_yardoc/
 /coverage/
-/doc/
 /pkg/
 /spec/reports/
 /tmp/
--- a/doc/net-http/examples.rdoc
+++ b/doc/net-http/examples.rdoc
@@ -1,0 +1,30 @@
+Examples here assume that <tt>net/http</tt> has been required
+(which also requires +uri+):
+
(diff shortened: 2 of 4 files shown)


BurdetteLamar:
This will also require a change over at ruby/ruby/tool/sync_default_gems.rb.

peterzhu2118 (approved):

BurdetteLamar:
> This will also require a change over at ruby/ruby/tool/sync_default_gems.rb.

@peterzhu2118: Does that work need to be synced with this?  Or with a merge from ruby/net-http into ruby/ruby?

peterzhu2118:
I think it would automatically sync. But I'm not 100% sure. Let's try, if it fails to sync I'll look into it and manually sync by hand.

peterzhu2118:
Seemed to have properly sync'd: ruby/ruby@909ea6b60d658b2a1c4a53be81b09642dcf86d4b

BurdetteLamar:
> Seemed to have properly sync'd: [ruby/ruby@909ea6b](https://github.com/ruby/ruby/commit/909ea6b60d658b2a1c4a53be81b09642dcf86d4b)

Thanks.

PR #79, opened 2022-11-19
[DOC] Enhanced RDoc for Net::HTTPHeader

BurdetteLamar:
Expands class doc.  Treats methods:
- #[]
- #[]=
- #add_field
- #get_fields
- #fetch
- #each_header
- #each_name
- #each_capitalized_name

### DIFF
--- a/lib/net/http/header.rb
+++ b/lib/net/http/header.rb
@@ -38,7 +38,7 @@
 #
 # === Field Values
 #
-# A field value may be returned as an array or as a string:
+# A field value may be returned as an array of strings or as a string:
 #
 # - These methods return field values as arrays:
 #
@@ -145,15 +145,6 @@
 # - #each_header: Passes each field name/value pair to the block.
 # - #each_name: Passes each field name to the block.
 # - #each_value: Passes each field value to the block.
-#
-# == Hash of Headers
-#
-# Each of the iterator methods above returns a hash of the headers of +self+.
-# Note that modifying that returned hash actually modifies the headers:
-#
-#   h = req.each_header {|k, v| }
-#   h.clear
-#   req.to_hash # => {}
 #
 module Net::HTTPHeader
 
@@ -280,10 +271,12 @@ module Net::HTTPHeader
   end
 
   # :call-seq
-  #   fetch(key, default_val = nil) {|key| ... }
+  #   fetch(key, default_val = nil) {|key| ... } -> object
+  #   fetch(key, default_val = nil) value or default_val
   #
   # With a block, returns the string value for +key+ if it exists;
   # otherwise returns the value of the block;
+  # ignores the +default_val+;
   # see {Fields}[rdoc-ref:Net::HTTPHeader@Fields]:
   #
   #   req = Net::HTTP::Get.new(uri)
@@ -304,9 +297,7 @@ module Net::HTTPHeader
     a.kind_of?(Array) ? a.join(', ') : a
   end
 
-  # Calls the block with each key/value pair;
-  # returns the {hash of headers}[rdoc-ref:Net::HTTPHeader@Hash+of+Headers];
-  # see {Fields}[rdoc-ref:Net::HTTPHeader@Fields]:
+  # Calls the block with each key/value pair:
   #
   #   req = Net::HTTP::Get.new(uri)
   #   req.each_header {|key, value| p [key, value] }
@@ -330,9 +321,7 @@ module Net::HTTPHeader
 
   alias each each_header
 
-  # Calls the block with each field key;
-  # returns the {hash of headers}[rdoc-ref:Net::HTTPHeader@Hash+of+Headers];
-  # see {Fields}[rdoc-ref:Net::HTTPHeader@Fields]:
+  # Calls the block with each field key:
   #
   #   req = Net::HTTP::Get.new(uri)
   #   req.each_key {|key| p key }
@@ -354,9 +343,7 @@ module Net::HTTPHeader
 
   alias each_key each_name
 
-  # Calls the block with each capitalized field name;
-  # returns the {hash of headers}[rdoc-ref:Net::HTTPHeader@Hash+of+Headers];
-  # see {Fields}[rdoc-ref:Net::HTTPHeader@Fields]:
+  # Calls the block with each capitalized field name:
   #
   #   req = Net::HTTP::Get.new(uri)
   #   req.each_capitalized_name {|key| p key }
--- a/lib/net/http/header.rb
+++ b/lib/net/http/header.rb
@@ -272,7 +272,7 @@ module Net::HTTPHeader
 
   # :call-seq
   #   fetch(key, default_val = nil) {|key| ... } -> object
-  #   fetch(key, default_val = nil) value or default_val
+  #   fetch(key, default_val = nil) -> value or default_val
   #
   # With a block, returns the string value for +key+ if it exists;
   # otherwise returns the value of the block;
--- a/lib/net/http/header.rb
+++ b/lib/net/http/header.rb
@@ -513,7 +513,7 @@ module Net::HTTPHeader
     r
   end
 
-  n range= set_range
+  alias range= set_range
 
   # Returns an Integer object which represents the HTTP Content-Length:
   # header field, or +nil+ if that field was not provided.
--- a/lib/net/http/header.rb
+++ b/lib/net/http/header.rb
@@ -20,26 +20,21 @@
 #
 # A field key may be:
 #
-# - A string: Key <tt>'Foo'</tt> is treated as if it were
-#   <tt>'Foo'.downcase</tt>;  i.e., <tt>'foo'</tt>.
-# - A symbol: Key <tt>:Foo</tt> is treated as if it were
-#   <tt>:Foo.to_s.downcase</tt>;  i.e., <tt>'foo'</tt>.
+# - A string: Key <tt>'Accept'</tt> is treated as if it were
+#   <tt>'Accept'.downcase</tt>;  i.e., <tt>'accept'</tt>.
+# - A symbol: Key <tt>:Accept</tt> is treated as if it were
+#   <tt>:Accept.to_s.downcase</tt>;  i.e., <tt>'accept'</tt>.
 #
 # Examples:
 #
-#   req = Net::HTTP.get(uri)
-#   req[:Accept]  # => "*/*"
+#   req = Net::HTTP::Get.new(uri)
+#   req[:accept]  # => "*/*"
 #   req['Accept'] # => "*/*"
 #   req['ACCEPT'] # => "*/*"
 #
 #   req['accept'] = 'text/html'
-#   req['accept'] # => "text/html"
-#   req[:Accept] = 'text/html'
-#   req['accept'] # => "text/html"
-#   req['Accept'] = 'application/json'
-#   req['accept'] # => "application/json"
-#   req['ACCEPT'] = 'text/plain'
-#   req['accept'] # => "text/plain"
+#   req[:accept] = 'text/html'
+#   req['ACCEPT'] = 'text/html'
 #
 # === Field Values
 #
@@ -73,15 +68,15 @@
 #
 # - \String:
 #
-#     req[:foo] = 'bar'
-#     req[:foo]            # => "bar"
-#     req.get_fields(:foo) # => ["bar"]
+#     req['Accept'] = 'text/html' # => "text/html"
+#     req['Accept']               # => "text/html"
+#     req.get_fields('Accept')    # => ["text/html"]
 #
 # - \Symbol:
 #
-#     req[:foo] = :bar
-#     req[:foo]            # => "bar"
-#     req.get_fields(:foo) # => ["bar"]
+#     req['Accept'] = :text    # => :text
+#     req['Accept']            # => "text"
+#     req.get_fields('Accept') # => ["text"]
 #
 # - Simple array:
 #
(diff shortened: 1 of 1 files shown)


lib/net/http/header.rb
```diff
@@ -1,16 +1,159 @@
...
+#
+# == Convenience Methods
+#
+# Various convenience methods retrieve values, set values, query values,
+# set form values, or iterate over fields.
+#
+# === Getters
+#
+# - #[]: Returns the string value for the given field.
+# - #content_length: Returns the integer value of field +:content-length+.
```
peterzhu2118, line 116: Although the header keys are not case sensitive, it's convention to use capitalize the first character of every word.

Also, you can't have dashes in Ruby symbols without quoting it.

```suggestion
# - #content_length: Returns the integer value of field +:'Content-Length'+.
```

Or maybe it's simpler to just use a string:

```suggestion
# - #content_length: Returns the integer value of field +'Content-Length'+.
```

lib/net/http/header.rb
```diff
@@ -1,16 +1,159 @@
...
+# - #content_range: Returns the Range value of field +:content-range+.
+# - #content_type: Returns the string value of field +:content-type+.
+# - #main_type: Returns first part of the string value of field +:content-type+.
+# - #sub_type: Returns second part of the string value of field +:content-type+.
+# - #range: Returns an array of Range objects, or +nil+.
```
peterzhu2118, line 121:
```suggestion
# - #range: Returns an array of Range objects of field +'Range'+, or +nil+.
```

lib/net/http/header.rb
```diff
@@ -1,16 +1,159 @@
...
+# - #range_length: Returns the integer length of the range given in field +:range+.
```
peterzhu2118, line 122: Looking at the documentation and source code, it looks at the `Content-Range` header (and not `Range`).

```suggestion
# - #range_length: Returns the integer length of the range given in field +'Content-Range'+.
```

lib/net/http/header.rb
```diff
@@ -1,16 +1,159 @@
...
+# - #type_params: Returns the string parameters for +:content-type+.
+#
+# === Setters
+#
+# - #[]=: Sets the string or array value for the given field.
+# - #basic_auth: Sets the string authorization header for +:Basic+ authorization.
```
peterzhu2118, line 128: It sets the `Authorization` header.

```suggestion
# - #basic_auth: Sets the string authorization header for +'Authorization'+.
```

lib/net/http/header.rb
```diff
@@ -1,16 +1,159 @@
...
+# - #content_length=: Sets the integer length for field +:content-length+.
+# - #content_type=: Sets the string value for field +:content-type+.
+# - #proxy_basic_auth: Set Proxy-Authorization: header for “Basic” authorization.
```
peterzhu2118, line 131:
```suggestion
# - #proxy_basic_auth: Sets the string authorization header for +'Proxy-Authorization'+.
```

lib/net/http/header.rb
```diff
@@ -1,16 +1,159 @@
...
+# - #range=: Sets the HTTP Range: header. Accepts either a Range object as a single argument, or a beginning index and a length from that index. Example:
```
peterzhu2118, line 132: I don't think we need to talk about the specifics of the method here.

```suggestion
# - #range=: Sets the value for field +'Range'+.
```

lib/net/http/header.rb
```diff
@@ -1,16 +1,159 @@
...
+#
+# - #chunked?: Returns whether field +:transfer-encoding+ is set to <tt>'chunked'</tt>.
+# - #connection_close?: Returns whether field +:connection+ is set to <tt>'close'</tt>.
+# - #connection_keep_alive?: Returns whether field +:connection+ is set to <tt>'keep-alive'</tt>.
+# - #key?: Returns whether a given field exists.
+#
+# === Form Setters
+#
+# - #set_form: Sets an HTML form data set.
+# - #set_form_data: Set header fields and a body from HTML form data.
```
peterzhu2118, line 144: To prevent linking to the `Set` class.

```suggestion
# - #set_form_data: \Set header fields and a body from HTML form data.
```

lib/net/http/header.rb
```diff
@@ -33,14 +176,30 @@ def size   #:nodoc: obsolete
 
   alias length size   #:nodoc: obsolete
 
-  # Returns the header field corresponding to the case-insensitive key.
-  # For example, a key of "Content-Type" might return "text/html"
+  # Returns the string field value for the case-insensitive field +key+,
+  # or +nil+ if there is no such key;
+  # see {Fields}[rdoc-ref:Net::HTTPHeader@Fields]:
+  #
+  #   req = Net::HTTP::Get.new(uri)
```
peterzhu2118, line 183: I feel like if you want to highlight the case-insensitiveness of this method, it would be useful to have a few examples of that here rather than in the documentation for `Net::HTTP::Header`.

lib/net/http/header.rb
```diff
@@ -101,41 +258,63 @@ def add_field(key, val)
...
     return nil unless @header[stringified_downcased_key]
     @header[stringified_downcased_key].dup
   end
 
-  # Returns the header field corresponding to the case-insensitive key.
-  # Returns the default value +args+, or the result of the block, or
-  # raises an IndexError if there's no header field named +key+
-  # See Hash#fetch
+  # :call-seq
+  #   fetch(key, default_val = nil) {|key| ... }
```
peterzhu2118, line 279: I think we should have a call-seq without the block to show that the block is optional.

lib/net/http/header.rb
```diff
@@ -101,41 +258,63 @@ def add_field(key, val)
...
+  #   req.fetch(:foo) {|key| key.capitalize }    # => "bar"
+  #   req.fetch(:nosuch) {|key| key.capitalize } # => "Nosuch"
+  #
+  # With no block, returns the string value for +key+ if it exists;
+  # otherwise, returns +default_val+ if it was given;
+  # otherwise raises an exception:
+  #
+  #   req.fetch(:foo)          # => "bar"
+  #   req.fetch(:nosuch, :baz) # => :baz
+  #   req.fetch(:nosuch)       # Raises IndexError.
```
peterzhu2118, line 296: It raises KeyError.

```suggestion
  #   req.fetch(:nosuch)       # Raises KeyError.
```

lib/net/http/header.rb
```diff
@@ -101,41 +258,63 @@ def add_field(key, val)
...
+  #
   def fetch(key, *args, &block)   #:yield: +key+
     a = @header.fetch(key.downcase.to_s, *args, &block)
     a.kind_of?(Array) ? a.join(', ') : a
   end
 
-  # Iterates through the header names and values, passing in the name
-  # and value to the code block supplied.
+  # Calls the block with each key/value pair;
+  # returns the value of #to_hash;
```
peterzhu2118, line 304: Technically, it doesn't return the same thing as `#to_hash` since `#to_hash` duplicates the internal hash (thus preventing modifications to the returned hash to affect the internal hash, and vice versa), whereas this one directly returns the internal hash. Therefore, I'm not sure if it's a good idea to advertise to readers that the return value of this method is the same as `#to_hash`.

lib/net/http/header.rb
```diff
@@ -101,41 +258,63 @@ def add_field(key, val)
...
-  # Example:
+  # Output:
   #
-  #     response.header.each_header {|key,value| puts "#{key} = #{value}" }
+  #   ["accept-encoding", "gzip;q=1.0,deflate;q=0.6,identity;q=0.3"]
+  #   ["accept", "*/*"]
+  #   ["user-agent", "Ruby"]
+  #   ["host", "jsonplaceholder.typicode.com"]
   #
+  # Returns an enumerator if no block is given.
```
peterzhu2118, line 317: Should we say that `each` is an alias of `each_header`?

lib/net/http/header.rb
```diff
@@ -145,23 +324,46 @@ def each_header   #:yield: +key+, +value+
...
+  # Output:
+  #
+  #   "accept-encoding"
+  #   "accept"
+  #   "user-agent"
+  #   "host"
   #
   # Returns an enumerator if no block is given.
+  #
+  # Net::HTTPHeader#each_name is an alias for Net::HTTPHeader#each_key.
```
peterzhu2118, line 343: I don't think we need to use the full module name when we're in that module.

```suggestion
  # #each_name is an alias for #each_key.
```

(Also technically, it's the other way around: `#each_key` is an alias for `#each_name`).

BurdetteLamar: I'm aware, but am trying to use one term -- key -- as much as possible, to avoid confusion between key and name.

lib/net/http/header.rb
```diff
@@ -1,16 +1,159 @@
...
+#   req['accept'] = 'text/html'
+#   req['accept'] # => "text/html"
+#   req[:Accept] = 'text/html'
+#   req['accept'] # => "text/html"
+#   req['Accept'] = 'application/json'
+#   req['accept'] # => "application/json"
+#   req['ACCEPT'] = 'text/plain'
+#   req['accept'] # => "text/plain"
+#
+# === Field Values
```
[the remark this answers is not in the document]
BurdetteLamar: So, no change requested here, right?

lib/net/http/header.rb
BurdetteLamar: > I don't think we need to use the full module name when we're in that module.

The doc guide is explicit on this point.  So, stet.

lib/net/http/header.rb
BurdetteLamar: There are many such places thoughout the core, where the call-seq shows the block and the text below says stuff like "Returns an enumerator if no block is given."  Should we amend the doc guide, one way or the other?

lib/net/http/header.rb
BurdetteLamar: I'm thinking to change all the keys to capitalized strings, with no lowercased and no symbols.  That's what people are used to seeing.  I don't think many care about the case-insensivity (though the description should still mention it).

lib/net/http/header.rb
```diff
@@ -1,16 +1,159 @@
...
+# A field key may be:
+#
+# - A string: Key <tt>'Foo'</tt> is treated as if it were
+#   <tt>'Foo'.downcase</tt>;  i.e., <tt>'foo'</tt>.
+# - A symbol: Key <tt>:Foo</tt> is treated as if it were
+#   <tt>:Foo.to_s.downcase</tt>;  i.e., <tt>'foo'</tt>.
+#
+# Examples:
+#
+#   req = Net::HTTP.get(uri)
```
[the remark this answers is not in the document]
BurdetteLamar: Thinko.  Should have called Get.new.  Fixed.

lib/net/http/header.rb
```diff
@@ -1,16 +1,159 @@
...
+#   req['ACCEPT'] # => "*/*"
+#
+#   req['accept'] = 'text/html'
+#   req['accept'] # => "text/html"
+#   req[:Accept] = 'text/html'
+#   req['accept'] # => "text/html"
+#   req['Accept'] = 'application/json'
+#   req['accept'] # => "application/json"
+#   req['ACCEPT'] = 'text/plain'
+#   req['accept'] # => "text/plain"
```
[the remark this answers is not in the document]
BurdetteLamar: Culled.

lib/net/http/header.rb
BurdetteLamar: Changed all to capitalized and quoted (except  the examples showing case and symbol).

lib/net/http/header.rb

lib/net/http/header.rb

lib/net/http/header.rb

lib/net/http/header.rb

lib/net/http/header.rb

lib/net/http/header.rb
BurdetteLamar: Here and one other: should have been Sets consistent with others).

lib/net/http/header.rb

lib/net/http/header.rb
BurdetteLamar: I've added a link to a warning (3 places).

lib/net/http/header.rb
BurdetteLamar: All aliases now noted.

lib/net/http/header.rb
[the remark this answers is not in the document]
peterzhu2118: Well this section implicitly shows how arrays and hashes are converted to strings, maybe we can move that to a separate section to highlight that behavior?

lib/net/http/header.rb
peterzhu2118: Well for example, here's the call-seq for Hash#each:

```
each {|key, value| ... } → self
each → new_enumerator
```

And here's the one for Hash#fetch (which is what this is calling internally):

```
fetch(key) → object
fetch(key, default_value) → object
fetch(key) {|key| ... } → object
```

So I think we should have a similar call-seq as `Hash#fetch`.

lib/net/http/header.rb
peterzhu2118: I think we should just omit it. They shouldn't be using the return value anyways, it's just a side-effect of Ruby returning the last thing in the method. Users shouldn't depend on it.

lib/net/http/header.rb
BurdetteLamar: Done.

lib/net/http/header.rb
BurdetteLamar: Done.

lib/net/http/header.rb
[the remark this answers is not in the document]
BurdetteLamar: I don't know what you mean.   The example field values has give segments.  That's not explicit enough?

BurdetteLamar:
@peterzhu2118 ready for your review.

lib/net/http/header.rb
```diff
@@ -101,41 +253,65 @@ def add_field(key, val)
...
     @header[stringified_downcased_key].dup
   end
 
-  # Returns the header field corresponding to the case-insensitive key.
-  # Returns the default value +args+, or the result of the block, or
-  # raises an IndexError if there's no header field named +key+
-  # See Hash#fetch
+  # :call-seq
+  #   fetch(key, default_val = nil) {|key| ... } -> object
+  #   fetch(key, default_val = nil) value or default_val
```
peterzhu2118, line 275:
```suggestion
  #   fetch(key, default_val = nil) -> value or default_val
```

lib/net/http/header.rb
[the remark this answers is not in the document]
peterzhu2118: What I mean is that it was a surprise for me that we can have values of arrays and hashes since HTTP headers don't allow these types (it's a non-standard thing Ruby is doing). How it converts the arrays and hashes to string is also surprising to me (even more so for hashes, since it creates an array of keys and values). I think it's worth highlighting since IMO it's very surprising behavior.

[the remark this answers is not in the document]
BurdetteLamar: I was surprised, too.  But I think that the separate examples for arrays, hashes, and mixtures is doing that work.  You may have to show me how you think this should be.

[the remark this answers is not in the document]
peterzhu2118: If you think the explanation is sufficient, then I'm ok with it 👍

peterzhu2118 (approved):

hsbt:
@BurdetteLamar Please do "squash and merge" if you use the same commit message. I still wonder why you don't use it. I'm tired of saying it over and over again.

BurdetteLamar:
> @BurdetteLamar Please do "squash and merge" if you use the same commit message. I still wonder why you don't use it. I'm tired of saying it over and over again.

I thought I had done so (and certainly meant to do so).  Is that not the default for the select list in the merge button?  If not, can I make it the default (to prevent my making this error in the future)?

hsbt:
I can change default strategy. Is there the reason we change the default configuration for only your mistake?

I also wonder why you still ignore "Please write the meaningful message". You must learn how to use Git and version control of software before the Ruby development.

peterzhu2118:
Make sure to use `Squash and merge` (found in the dropdown menu) before merging. That will prevent this kind of issue in the future.

I don't think there's a feature in GitHub to select a default merge method for your account.

Screenshot 2022-11-23 at 5 11 58 PM

BurdetteLamar:
> I can change default strategy. Is there the reason we change the default configuration for only your mistake?
> 
> I also wonder why you still ignore "Please write the meaningful message". You must learn how to use Git and version control of software before the Ruby development.

It might be better if I don't merge at all, but instead ask my wonderful reviewers (like @peterzhu2118, @jeremyevans, and others) to not only review/approve, but also to merge.

It isn't easy, being old.  (I'll be 80 in a few weeks.)

peterzhu2118:
If you don't feel comfortable merging, I don't mind doing it for you.

BurdetteLamar:
> If you don't feel comfortable merging, I don't mind doing it for you.

Thanks, @peterzhu2118!

PR #83, opened 2022-11-29
[DOC] Enhanced RDoc for Net::HTTPHeader

BurdetteLamar:
PR to emphasize that setters are mostly about requests, while getters are mostly about responses:
- Moved setter sections before getter sections; you do set/request before response/get.
- Added introductory remark to setters section.
- Added #add_field to setters section.
- Added introductory remark to getters section.
- Added #fetch to getters section.
- Fine-tuned distinction between string field value and array field value (a few places).
- Changed examples for getters to use a response object instead of a request object.

### DIFF
--- a/lib/net/http/header.rb
+++ b/lib/net/http/header.rb
@@ -128,12 +128,11 @@
 #
 # === Setters
 #
-# \Method #[]= can set any field, but does little to validate
-# to validate the new value;
-# most of the other setter methods provides some validation:
+# \Method #[]= can set any field, but does little to validate the new value;
+# some of the other setter methods provide some validation:
 #
 # - #[]=: Sets the string or array value for the given key.
-# - #add_field: Creates or adds to the array value for the given kry.
+# - #add_field: Creates or adds to the array value for the given key.
 # - #basic_auth: Sets the string authorization header for <tt>'Authorization'</tt>.
 # - #content_length=: Sets the integer length for field <tt>'Content-Length</tt>.
 # - #content_type=: Sets the string value for field <tt>'Content-Type'</tt>.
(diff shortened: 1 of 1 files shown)


lib/net/http/header.rb
```diff
@@ -126,46 +126,59 @@
 # Various convenience methods retrieve values, set values, query values,
 # set form values, or iterate over fields.
 #
+# === Setters
+#
+# \Method #[]= can set any field, but does little to validate
+# to validate the new value;
```
peterzhu2118, line 132:
```suggestion
# \Method #[]= can set any field, but does little to validate
# the new value;
```

lib/net/http/header.rb
```diff
@@ -126,46 +126,59 @@
...
+# most of the other setter methods provides some validation:
+#
+# - #[]=: Sets the string or array value for the given key.
+# - #add_field: Creates or adds to the array value for the given kry.
```
peterzhu2118, line 136:
```suggestion
# - #add_field: Creates or adds to the array value for the given key.
```

peterzhu2118, line 136: Technically, `#add_field` is also like `#[]=` in that it doesn't validate the value.

lib/net/http/header.rb
```diff
@@ -126,46 +126,59 @@
...
+# but always as a string;
+# most of the other getter methods return something different
+# from the simple string value:
+#
+# - #[]: Returns the string field value for the given key.
 # - #content_length: Returns the integer value of field <tt>'Content-Length'</tt>.
 # - #content_range: Returns the Range value of field <tt>'Content-Range'</tt>.
 # - #content_type: Returns the string value of field <tt>'Content-Type'</tt>.
+# - #fetch: Returns the string field value for the given key,
+#   or passes it to the block, if given.
```
peterzhu2118, line 160: `fetch` returns the default value or the value of the block if it doesn't exist (and doesn't pass the value to the block). I couldn't think of a good way to explain that in the short sentence, so I just copied the sentence from `Hash#fetch`.

```suggestion
# - #fetch: Returns the string field value for the given key.
```

lib/net/http/header.rb
BurdetteLamar: Fixed.

lib/net/http/header.rb
BurdetteLamar: Fixed.

BurdetteLamar: I've weakened the validation claim.

peterzhu2118 (approved):

PR #87, opened 2022-12-08
don't invoke response block more than once due to retry

### LINKED ISSUE #86: retry in streaming response causes corrupt downloads
If a socket error occurs while performing a streaming download via the response block provided to transport_request, a retry may cause the response block may be invoked a second time.

For code patterns like this one:

```
  open('some_file', 'wb') do |io|
    http.request(...) do |response|
      response.read_body do |chunk|
        io.write chunk
      end
    end
  end
```

The output file will be corrupted since some portion of its data near the beginning will be duplicated.

jstanley0:
If a socket error occurs while performing a streaming download via the response block provided to transport_request, avoid calling the response block again as this would result in duplicate data received by the client.

fixes #86

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -1634,7 +1634,10 @@ module Net   #:nodoc:
           res
         }
         res.reading_body(@socket, req.response_body_permitted?) {
-          yield res if block_given?
+          if block_given?
+            count = max_retries # Don't restart in the middle of a download
+            yield res
+          end
         }
       rescue Net::OpenTimeout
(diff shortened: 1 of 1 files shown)


jeremyevans (approved):
It would be good if this added tests, but regardless I think it should be committed.  This fixes Ruby Bug #11526.

nobu (changes requested):
A test would be required.

nurse:
@jeremyevans You can merge this if you provide a test

PR #88, opened 2022-12-11
[DOC] Enhanced RDoc for HTTPHeader

BurdetteLamar:
Treats:
- #content_length
- #content_length=
- #chunked?
- #content_range
- #range_length
- #content_type
- #main_type
- #sub_type
- #type_params
- #set_content_type

### DIFF
--- a/lib/net/http/header.rb
+++ b/lib/net/http/header.rb
@@ -381,11 +381,11 @@ module Net::HTTPHeader
   #
   # Output:
   #
-  # "content-type"
-  # "connection"
-  # "cache-control"
-  # "cf-cache-status"
-  # "cf-ray"
+  #   "content-type"
+  #   "connection"
+  #   "cache-control"
+  #   "cf-cache-status"
+  #   "cf-ray"
   #
   # Returns an enumerator if no block is given.
   #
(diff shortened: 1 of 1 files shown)


BurdetteLamar:
Some of the things we're linking to over at Wikipdia.org already had ids or anchors, some didn't.   I've been adding some ids and anchors as we go.

lib/net/http/header.rb
```diff
@@ -610,15 +610,40 @@ def set_range(r, e = nil)
...
 
-  # Returns an Integer object which represents the HTTP Content-Length:
-  # header field, or +nil+ if that field was not provided.
+  # Returns the value of field <tt>'Content-Length'</tt> as an integer,
+  # or +nil+ if there is no such field;
+  # see {Content-Length request header}[https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#content-length-request-header]:
+  #
+  #   res = Net::HTTP.start(hostname) do |http|
+  #     http.get('/nosuch/1')
+  #   end # => #<Net::HTTPNotFound 404 Not Found readbody=true>
```
peterzhu2118, line 619: Wouldn't it be easier to use `Net::HTTP.get` here to make it one line?

lib/net/http/header.rb
```diff
@@ -627,20 +652,35 @@ def content_length=(len)
     @header['content-length'] = [len.to_i.to_s]
   end
 
-  # Returns "true" if the "transfer-encoding" header is present and
-  # set to "chunked".  This is an HTTP/1.1 feature, allowing
-  # the content to be sent in "chunks" without at the outset
-  # stating the entire content length.
+  # Returns +true+ if field <tt>'Transfer-Encoding'
```
peterzhu2118, line 655:
```suggestion
  # Returns +true+ if field <tt>'Transfer-Encoding'</tt>
```

lib/net/http/header.rb
BurdetteLamar: I wanted to emphasize the use of start, on the theory that many users will have more than one request to make and should be aware of it.  There are a number of these (16?) in this file.  If one changes, should all?

BurdetteLamar: So, change all?

lib/net/http/header.rb
```diff
@@ -207,9 +207,7 @@ def size   #:nodoc: obsolete
   # or +nil+ if there is no such key;
   # see {Fields}[rdoc-ref:Net::HTTPHeader@Fields]:
   #
-  #   res = Net::HTTP.start(hostname) do |http|
-  #     http.get('/todos/1')
-  #   end
+  #   res = Net::HTTP.get(hostname, '/todos/1')
```
peterzhu2118, line 210: Oops, I made a mistake, `Net::HTTP.get` returns a string, we should use `Net::HTTP.HTTP.get_response` to get a `Net::HTTPResponse `.

BurdetteLamar: Et mea culpa.  I should have checked at least one complete example.

peterzhu2118 (approved):
Looks good!

PR #89, opened 2022-12-13

BurdetteLamar:
Treats:
- #inspect
- #set_debug_output
- #response_body_encoding=
- #ipaddr
- #ipaddr=
- #max_retries=

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -864,6 +864,11 @@ module Net   #:nodoc:
       end
     end
 
+    # Returns a string representation of +self+:
+    #
+    #   Net::HTTP.new(hostname).inspect
+    #   # => "#<Net::HTTP jsonplaceholder.typicode.com:80 open=false>"
+    #
     def inspect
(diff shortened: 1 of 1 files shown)


lib/net/http.rb
```diff
@@ -942,12 +1028,19 @@ def ipaddr=(addr)
...
-    # Should be a non-negative integer number. Zero means no retries.
-    # The default value is 1.
+    # The initial value is 1.
+    #
+    # Argument +retries+ must be a non-negative numeric value:
+    #
+    #   http.max_retries = 2   # => 2
+    #   http.max_retries       # => 2
+    #   http.max_retries = 2.9 # => 2.9
+    #   http.max_retries       # => 2
```
peterzhu2118, line 1042: I know that it's converting it to an integer, but this is a really weird example, let's not encourage people to set floats for these kinds of values 😂

BurdetteLamar: Two lines removed from examples.

peterzhu2118 (approved):

PR #90, opened 2022-12-14
[DOC] Correct formatting in header.rb

BurdetteLamar:


lib/net/http/header.rb
```diff
@@ -301,7 +301,7 @@ def get_fields(key)
     @header[stringified_downcased_key].dup
   end
 
-  # :call-seq
+  # :call-seq:
```
peterzhu2118, line 304: Isn't it `call-seq:` (colon only on the right side)?

BurdetteLamar: The [RDoc markup reference](https://docs.ruby-lang.org/en/master/RDoc/MarkupReference.html#class-RDoc::MarkupReference-label-Directives+for+Method+Documentation) has it as `:call-seq:`, but Ruby-dev custom uses only the trailing colon.

Would you kindly fix here (there may be more than one in this file).  I'll PR for other files separately.

peterzhu2118 (approved):

PR #101, opened 2023-01-08
Test against Ruby 3.2.0

gfvcastro:


amatsuda:
@gfvcastro Please do not the existing `'3.0'`. As shown in this CI result:
https://github.com/ruby/net-http/actions/runs/3864905252/jobs/6588048949
GH Actions' YAML parser parses `3.0` as integer `3`, then it ends up running the tests on "the latest released CRuby 3.x", i.e. 3.2.0 as of today.

See https://github.com/actions/runner/issues/849 for more details.

PR #103, opened 2023-01-10
[DOC] Enhanced RDoc for set_form

BurdetteLamar:


BurdetteLamar:
Would it be best to leave this one alone?

lib/net/http/header.rb
```diff
@@ -806,6 +806,107 @@ def set_form_data(params, sep = '&')
 
   alias form_data= set_form_data
 
+  # Stores form data to be included in the body of a +POST+ or +PUT+ request.
+  # When the request is sent, the data is URL-encoded
+  # (see {URI#encode_www_form}[https://docs.ruby-lang.org/en/master/URI.html#method-c-encode_www_form])
```
peterzhu2118, line 811: This is only true if `enctype` is `application/x-www-form-urlencoded`. If `enctype` is `multipart/form-data`, then it's encoded differently. For example, here I'll directly use the private method `encode_multipart_form_data` to encode a hash as `multipart/form-data`:


```ruby
uri = URI('http://example.com')
req = Net::HTTP::Get.new(uri)
buf = StringIO.new
req.send(:encode_multipart_form_data, buf, {"testing" => "hello"}, {})
puts buf.string
# --VW1nIsu7UHRJGW-ROHq6aeAadUQ87H4CsUoCvcyYSBK6qxQu43nnVg
# Content-Disposition: form-data; name="testing"
#
# hello
# --VW1nIsu7UHRJGW-ROHq6aeAadUQ87H4CsUoCvcyYSBK6qxQu43nnVg--
```

lib/net/http/header.rb
```diff
@@ -806,6 +806,107 @@ def set_form_data(params, sep = '&')
...
+  # and included in the request body.
+  #
+  # Note: There is no method to retrieve the form data directly;
+  # here we define a convenience method to do just that
+  # (its details are unimportant):
+  #
+  #   def get_form(req)
+  #     YAML.unsafe_load(req.to_yaml.lines[1..].join)['body_data']
+  #   end
```
peterzhu2118, line 820: I'm not a fan of this. We shouldn't be trying to find hacky ways to get the internal state of the object.

For `application/x-www-form-urlencoded`, we can just let the reader refer to the documentation of `URI.encode_www_form`.

For `multipart/form-data`, ... I'm not too sure what the best way to show examples of that is.

lib/net/http/header.rb
```diff
@@ -806,6 +806,107 @@ def set_form_data(params, sep = '&')
...
+  #
+  # Any input referred to here as a _string_
+  # may actually be a
+  # {string-convertible object}[https://docs.ruby-lang.org/en/master/implicit_conversion_rdoc.html#label-String-Convertible+Objects].
```
peterzhu2118, line 824: This is confusing to me. When I read the word "string", it's difficult for me to understand whether you mean string or a string-convertable object.

BurdetteLamar: Maybe better to just say 'string' and omit the note?

peterzhu2118: Yeah I think that would be less confusing.

lib/net/http/header.rb
BurdetteLamar: Removed.

lib/net/http/header.rb
BurdetteLamar: Removed.

lib/net/http/header.rb
BurdetteLamar: Removed.

BurdetteLamar:
I'm thinking that it might be best to restrict the doc for this method to just setting the values for `params`, `enc`, and `formopt`, and not getting into what happens when the request is to be sent.  The latter could be treated in the doc for POST and PUT.  Thoughts?

BurdetteLamar:
I think I'm in over my head with the documentation for Net::HTTPHeader#set_form.  I'd like to close this PR.

peterzhu2118:
Let me know if there's anything I can help with. If you think you don't want to work on this anymore, it's your call to close this.

BurdetteLamar:
@peterzhu2118, ready for review.

lib/net/http/header.rb
```diff
@@ -806,56 +806,107 @@ def set_form_data(params, sep = '&')
...
+  #
+  #     req.set_form([['foo'], ['bar'], ['baz']])
+  #     # String to be included in request body: 'foo&bar&baz'
+  #
+  # - Two strings:
+  #
+  #     req.set_form([%w[foo 0], %w[bar 1], %w[baz 2]])
+  #     # String to be include in request body: 'foo=0&bar=1&baz=2'
+  #
+  # - A string (ignored) and an IO stream opened for reading:
```
peterzhu2118, line 845: 1. This only applies for `multipart/form-data` and not `application/x-www-form-urlencoded`. For `application/x-www-form-urlencoded` it uses `URI.encode_www_form` which doesn't properly encode the IO stream:

```ruby
URI.encode_www_form([["foo", StringIO.new("foobar")]])
# => "foo=%23%3CStringIO%3A0x00000001060bdd88%3E"
```

  Whereas `'multipart/form-data'` does use the IO:

```ruby
io = StringIO.new
req.send(:encode_multipart_form_data, io, [["foo", StringIO.new("foobar")]], {})
puts io.string
# =>
# --KCyJx5jGk4z7PliS1mddX-fzCUfusQhEdg_EBzRp2SFxg17wzWP5Cw
# Content-Disposition: form-data; name="foo"
#
# foobar
# --KCyJx5jGk4z7PliS1mddX-fzCUfusQhEdg_EBzRp2SFxg17wzWP5Cw--
```

2. It does not ignore the first string. It uses it for the `name` field.

lib/net/http/header.rb
```diff
@@ -806,56 +806,107 @@ def set_form_data(params, sep = '&')
...
-  # you should confirm that the server supports HTTP/1.1 before using
-  # chunked encoding.
+  #   # IO-valued field.
+  #   File.write('t.tmp', 'Ruby is cool.')
+  #   file = File.open('t.tmp')
+  #   req.set_form({not_used: file})
   #
-  # Example:
+  #   # Mixture of fields.
+  #   req.set_form({foo: nil, bar: 1, not_used: file})
```
peterzhu2118, line 893: These examples use a symbol as the key of the hash, this doesn't work for `multipart/form-data`.

```ruby
req.send(:encode_multipart_form_data, io, {foo: "bar"}, {})
# => undefined method `gsub' for :foo:Symbol (NoMethodError)
```

lib/net/http/header.rb
```diff
@@ -806,56 +806,107 @@ def set_form_data(params, sep = '&')
...
-  #    )
+  # Optional argument +formopt+ is a hash of options that may include
+  # the following entries:
   #
-  # See also RFC 2388, RFC 2616, HTML 4.01, and HTML5
+  # - +:boundary+: The value is the boundary string for the multipart message.
+  #   If not given, the boundary is a random string.
+  #   See {Boundary}[https://www.rfc-editor.org/rfc/rfc7578#section-4.1].
+  # - +:charset+: Value is the character set for the form submission.
+  #   Field names and values of non-file fields should be encoded with this charset.
```
peterzhu2118, line 909: I think these are only applicable for `multipart/form-data`, we should note that.

lib/net/http/header.rb
BurdetteLamar: 1.  I've removed the "string to be included" remarks (four places), which I hope fixes this.
2. "string (ignored)" replaced by "string name" (two places).

lib/net/http/header.rb

lib/net/http/header.rb
BurdetteLamar: Noted.

lib/net/http/header.rb
```diff
@@ -806,56 +806,105 @@ def set_form_data(params, sep = '&')
...
+  #     req.set_form([['foo'], ['bar'], ['baz']])
+  #
+  # - Two strings:
+  #
+  #     req.set_form([%w[foo 0], %w[bar 1], %w[baz 2]])
+  #
+  # - A string name and an IO stream opened for reading:
+  #
+  #     File.write('t.tmp', 'Ruby is cool.')
+  #     file = File.open('t.tmp')
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed (2 places).

lib/net/http/header.rb
```diff
@@ -806,56 +806,105 @@ def set_form_data(params, sep = '&')
...
+  #     req.set_form([['not used', file]])
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed (5 places).

lib/net/http/header.rb
```diff
@@ -806,56 +806,105 @@ def set_form_data(params, sep = '&')
...
+  #
+  # When +params+ is a hash,
+  # each of its entries is a name/value pair that defines a field:
+  #
+  # - The name is a string.
+  # - The value may be:
+  #
+  #   - +nil+.
+  #   - Another string.
+  #   - An IO stream opened for reading.
```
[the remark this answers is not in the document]

lib/net/http/header.rb
```diff
@@ -806,56 +806,105 @@ def set_form_data(params, sep = '&')
...
   #
-  #    req.set_form([["f",
-  #                   File.open('/path/to/filename.bar'),
-  #                   {filename: "other-filename.foo"}
-  #                 ]],
-  #                 "multipart/form-data",
-  #    )
+  # Optional argument +formopt+ is a hash of options
+  # (applicable only when header <tt>'Content-Type'</tt>
+  # is <tt>'multipart/form-data'</tt>)
```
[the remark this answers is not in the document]

peterzhu2118 (commented):
A few more nitpicks, but otherwise I think it's good.

peterzhu2118 (approved):

lib/net/http/header.rb
```diff
@@ -806,56 +806,108 @@ def set_form_data(params, sep = '&')
...
+  # When +params+ is a hash,
+  # each of its entries is a name/value pair that defines a field:
+  #
+  # - The name is a string.
+  # - The value may be:
+  #
+  #   - +nil+.
+  #   - Another string.
+  #   - An IO stream opened for reading,
+  #     (only when argument +enctype+ (see below) is given as
```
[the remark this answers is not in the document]
BurdetteLamar: Grammarians differ on this point.  (I may count myself as a grammarian -- B.A., M.A. in English.)  Some say to go ahead and nest.  Others say to use inner square brackets;  still others say the square brackets must be reserved for editing remarks.  Almost all say to avoid if possible.  I've changed the inner punctuation to dashes.

PR #104, opened 2023-01-18
[DOC] Header doc

BurdetteLamar:
Treats:
- #basic_auth
- #proxy_basic_auth
- #connection_close?
- #connection_keep_alive?

### DIFF
--- a/lib/net/http/header.rb
+++ b/lib/net/http/header.rb
@@ -927,7 +927,8 @@ module Net::HTTPHeader
   # +account+ and +password+ strings:
   #
   #   req.basic_auth('my_account', 'my_password')
-  #   req['Authorization'] # => ["Basic bXlfYWNjb3VudDpteV9wYXNzd29yZA=="]
+  #   req['Authorization']
+  #   # => "Basic bXlfYWNjb3VudDpteV9wYXNzd29yZA=="
   #
   def basic_auth(account, password)
     @header['authorization'] = [basic_encode(account, password)]
@@ -937,7 +938,8 @@ module Net::HTTPHeader
   # +account+ and +password+ strings:
   #
   #   req.proxy_basic_auth('my_account', 'my_password')
-  #   req['Proxy-Authorization'] # => ["Basic bXlfYWNjb3VudDpteV9wYXNzd29yZA=="]
+  #   req['Proxy-Authorization']
+  #   # => "Basic bXlfYWNjb3VudDpteV9wYXNzd29yZA=="
   #
   def proxy_basic_auth(account, password)
     @header['proxy-authorization'] = [basic_encode(account, password)]
--- a/lib/net/http/header.rb
+++ b/lib/net/http/header.rb
@@ -927,7 +927,7 @@ module Net::HTTPHeader
   # +account+ and +password+ strings:
   #
   #   req.basic_auth('my_account', 'my_password')
-  #   # => ["Basic bXlfYWNjb3VudDpteV9wYXNzd29yZA=="]
+  #   req['Authorization'] # => ["Basic bXlfYWNjb3VudDpteV9wYXNzd29yZA=="]
   #
   def basic_auth(account, password)
     @header['authorization'] = [basic_encode(account, password)]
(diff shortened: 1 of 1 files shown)


BurdetteLamar:
Out-of-scope for this PR, but I don't think `#set_form_data`  and `#set_form` belong in Net::HTTPHeader.  They have much greater effects than merely modifying header entries.  Should they be moved to a more appropriate class/module?

peterzhu2118:
It might be a bit difficult to move these methods since they're tested as part of the spec. The class that includes the `Net::HTTPHeader` module is expected to have a `body` attribute that can be written to.

https://github.com/ruby/spec/blob/master/library/net/http/httpheader/shared/set_form_data.rb
https://github.com/ruby/spec/blob/master/library/net/http/httpheader/fixtures/classes.rb

lib/net/http/header.rb
```diff
@@ -923,12 +923,22 @@ def set_form(params, enctype='application/x-www-form-urlencoded', formopt={})
     end
   end
 
-  # Set the Authorization: header for "Basic" authorization.
+  # Sets header <tt>'Authorization'</tt> using the given
+  # +account+ and +password+ strings:
+  #
+  #   req.basic_auth('my_account', 'my_password')
+  #   # => ["Basic bXlfYWNjb3VudDpteV9wYXNzd29yZA=="]
```
peterzhu2118, line 930: I'm not a fan of depending on the return value of `#basic_auth` since it returns the internal state (array of a string) rather than the actual header value.

```suggestion
  #   req.basic_auth('my_account', 'my_password')
  #   req['Authorization'] # => "Basic bXlfYWNjb3VudDpteV9wYXNzd29yZA=="
```

lib/net/http/header.rb
```diff
@@ -923,12 +923,22 @@ def set_form(params, enctype='application/x-www-form-urlencoded', formopt={})
...
   def basic_auth(account, password)
     @header['authorization'] = [basic_encode(account, password)]
   end
 
-  # Set Proxy-Authorization: header for "Basic" authorization.
+  # Sets header <tt>'Proxy-Authorization'</tt> using the given
+  # +account+ and +password+ strings:
+  #
+  #   req.proxy_basic_auth('my_account', 'my_password')
+  #   # => ["Basic bXlfYWNjb3VudDpteV9wYXNzd29yZA=="]
```
peterzhu2118, line 940: Ditto here.

lib/net/http/header.rb

lib/net/http/header.rb

lib/net/http/header.rb
```diff
@@ -923,12 +923,22 @@ def set_form(params, enctype='application/x-www-form-urlencoded', formopt={})
     end
   end
 
-  # Set the Authorization: header for "Basic" authorization.
+  # Sets header <tt>'Authorization'</tt> using the given
+  # +account+ and +password+ strings:
+  #
+  #   req.basic_auth('my_account', 'my_password')
+  #   req['Authorization'] # => ["Basic bXlfYWNjb3VudDpteV9wYXNzd29yZA=="]
```
peterzhu2118, line 930: This returns a string, not an array.

```suggestion
  #   req['Authorization'] # => "Basic bXlfYWNjb3VudDpteV9wYXNzd29yZA=="
```

lib/net/http/header.rb
```diff
@@ -923,12 +923,22 @@ def set_form(params, enctype='application/x-www-form-urlencoded', formopt={})
...
   def basic_auth(account, password)
     @header['authorization'] = [basic_encode(account, password)]
   end
 
-  # Set Proxy-Authorization: header for "Basic" authorization.
+  # Sets header <tt>'Proxy-Authorization'</tt> using the given
+  # +account+ and +password+ strings:
+  #
+  #   req.proxy_basic_auth('my_account', 'my_password')
+  #   req['Proxy-Authorization'] # => ["Basic bXlfYWNjb3VudDpteV9wYXNzd29yZA=="]
```
peterzhu2118, line 940: Ditto.

lib/net/http/header.rb
BurdetteLamar: Fixed.  (This was slovenly work on my part.)

lib/net/http/header.rb

peterzhu2118 (approved):

PR #105, opened 2023-01-23
Initial approach to allowing SSL context option passing.

rcuhljr:
Hello!

This is really meant more to start a conversation since there's likely a lot of implications to this change I'm not aware of.

I've found that it would be very convenient to be able to pass in `options` to the underlying SSL Context, this provides support for flags like `OP_LEGACY_SERVER_CONNECT` and `OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION`. Which while dangerous do provide real utility. This was proposed in the [past](https://bugs.ruby-lang.org/issues/9450) however at the time it was for passing in the flags that block specific SSL versions. This proposal was closed when min/max version parameters were added. I think the underlying need is still there so I'm opening this up to get some feedback.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@
 /pkg/
 /spec/reports/
 /tmp/
+/.tool-versions
+/Gemfile.lock
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -658,6 +658,7 @@ module Net   #:nodoc:
     # - #open_timeout
     # - #read_timeout
     # - #ssl_timeout
+    # - #ssl_options
     # - #ssl_version
     # - +use_ssl+ (calls #use_ssl=)
     # - #verify_callback
@@ -1132,6 +1133,7 @@ module Net   #:nodoc:
       :@extra_chain_cert,
       :@key,
       :@ssl_timeout,
+      :@ssl_options,
       :@ssl_version,
       :@min_version,
       :@max_version,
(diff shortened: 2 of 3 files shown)


rcuhljr:
I'm super keen to be able to remove my projects patch on this someday, if anyone wants to weigh on this change.

htcarr3:
@rcuhljr, so if you set either the UnsafeLegacyServerConnect or UnsafeLegacyRenegotiation as options in the openssl configuration, does that not automatically get loaded into ruby? I'm running into this issue where, when I am working behind my corporate firewall we have proxy service that requires me to set these flags in order to use openssl. For some reason anything using ruby net/http still throws the renegotation error even though these options are set

jeffgran-dox:
I would love to see this get attention too. It would be nice to be able to set ssl options for http requests individually, and I also made a commit that would add the ability to set a default for all http requests: https://github.com/rcuhljr/net-http/pull/1

rcuhljr:
@htcarr3, @jeffgran-dox  Sorry this had really fallen off my radar over the back half of last year. As for setting in the overall configuration, I'm unsure but didn't pursue it since this is running on a web server for me I really don't want these settings as part of my overall SSL configuration for the session, just for some specific requests I'm making to 3rd party servers as health checks. Jeff thanks for the interest, and yeah having the extra flexibility of just doing it once would also be nice for situations that do want the same custom settings on all requests.

PR #106, opened 2023-01-23
[DOC] Clarify strategies

BurdetteLamar:
Help the user decide on a strategy (i.e., which request methods and classes to use).

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -43,8 +43,8 @@ module Net   #:nodoc:
   #
   # - If you will make only a few GET requests,
   #   consider using {OpenURI}[https://docs.ruby-lang.org/en/master/OpenURI.html].
-  # - If performance is _not_ important,
-  #   consider using various convenience methods in this class.
+  # - If you will make only a few requests of all kinds,
+  #   consider using the various singleton convenience methods in this class.
   #   Each of the following methods automatically starts and finishes
   #   a {session}[rdoc-ref:Net::HTTP@Sessions] that sends a single request:
   #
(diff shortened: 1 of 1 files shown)


lib/net/http.rb
```diff
@@ -34,67 +34,71 @@ class HTTPHeaderSyntaxError < StandardError; end
...
-  #
-  #   # Return string response body.
-  #   Net::HTTP.get(hostname, path, port = 80)
-  #   Net::HTTP.get(uri, headers = {}, port = 80)
+  # == Strategies
+  #
+  # - If you will make only a few GET requests,
+  #   consider using {OpenURI}[https://docs.ruby-lang.org/en/master/OpenURI.html].
+  # - If performance is _not_ important,
+  #   consider using various convenience methods in this class.
```
peterzhu2118, line 47: I'm not sure what this means. Why is using these methods bad for performance? If you're just sending a single request then it shouldn't be any slower right?

BurdetteLamar: Modified.

lib/net/http.rb
```diff
@@ -34,67 +34,71 @@ class HTTPHeaderSyntaxError < StandardError; end
...
+  #         # Session started automatically before block execution.
+  #         http.copy(path_or_uri, headers = {})
+  #         http.lock(path_or_uri, body, headers = {})
+  #         http.mkcol(path_or_uri, body = nil, headers = {})
+  #         http.move(path_or_uri, headers = {})
+  #         http.propfind(path_or_uri, body = nil, headers = {'Depth' => '0'})
+  #         http.proppatch(path_or_uri, body, headers = {})
+  #         http.unlock(path_or_uri, body, headers = {})
+  #         # Session finished automatically at block exit.
+  #       end
```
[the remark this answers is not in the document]
BurdetteLamar: Joined.

peterzhu2118 (approved):

PR #107, opened 2023-01-26
[DOC] Streamline examples

BurdetteLamar:
Main goal is for the examples to `include Net` so that `Net::HTTP` may be written as `HTTP`.  Also revises certain examples so that they may be executed without errors.

I hate to put in such a big PR, but having made the mods in examples.rdoc, I had to make appropriate changes in the three files that include it.

### DIFF
--- a/doc/net-http/examples.rdoc
+++ b/doc/net-http/examples.rdoc
@@ -1,7 +1,8 @@
 Examples here assume that <tt>net/http</tt> has been required
-(which also requires +uri+):
+(which also requires +uri+), and that <tt>Net</tt> has been included.
 
-  require 'net/http'
+  require 'net/http' # Defines Net::HTTP.
+  include Net        # Allows abbreviating Net::HTTP to HTTP.
 
 Many code examples here use these example websites:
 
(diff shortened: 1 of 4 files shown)


peterzhu2118 (changes requested):
I'm not a fan of including the `Net` module so that we can remove the `Net::` prefix. By including all of `Net`, every class/module under `Net` will be available in the level that you included it in (or in the top level if it was included in the top level). This makes it difficult to understand where a specific class/module comes from (e.g. most users know `Net::HTTPRequest` comes from this gem but will be confused when they see just `HTTPRequest`). It also pollutes the top level as every constant in the `Net` module will be part of the top level, thus increasing the probability of name collisions. This is also an abuse of includes, since including a module means that your code requires all of the code of the module being included to function.

BurdetteLamar:
> I'm not a fan of including the `Net` module so that we can remove the `Net::` prefix. By including all of `Net`, every class/module under `Net` will be available in the level that you included it in (or in the top level if it was included in the top level). This makes it difficult to understand where a specific class/module comes from (e.g. most users know `Net::HTTPRequest` comes from this gem but will be confused when they see just `HTTPRequest`). It also pollutes the top level as every constant in the `Net` module will be part of the top level, thus increasing the probability of name collisions. This is also an abuse of includes, since including a module means that your code requires all of the code of the module being included to function.

Got it.  I'll leave this up for a bit, in case others want to comment.  Meanwhile I'll collect the other changes (some of which are corrections) for a new PR.

PR #109, opened 2023-01-28
[DOC] Define path for examples

BurdetteLamar:


peterzhu2118 (commented):
I'm not sure what the purpose of this PR is. Why do we need a path for the URI?

BurdetteLamar:
> I'm not sure what the purpose of this PR is. Why do we need a path for the URI?

It will be needed for some changes to either header.rb or http.rb (I forget which).  I'll be replacing some schema-like examples with actual executable code, which will require a path.  I just put this up first b/c it's small and simple.

peterzhu2118:
I would prefer to have all the changes in a single PR so that I can see the motivation of this change. I can't see the benefits of this change so it's hard for me to give a meaningful review.

PR #112, opened 2023-01-31
[DOC] Revisions to README.md

BurdetteLamar:
Updates:
- Usage:  now points to the main class.
- Installation:  deleted (no longer needed).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,90 +1,14 @@
-# Net::HTTP
+# \Net::HTTP
 
-Net::HTTP provides a rich library which can be used to build HTTP
-user-agents.  For more details about HTTP see
-[RFC2616](http://www.ietf.org/rfc/rfc2616.txt).
-
-Net::HTTP is designed to work closely with URI.  URI::HTTP#host,
-URI::HTTP#port and URI::HTTP#request_uri are designed to work with
-Net::HTTP.
-
-If you are only performing a few GET requests you should try OpenURI.
-
-## Installation
-
-Add this line to your application's Gemfile:
-
-```ruby
-gem 'net-http'
-```
-
-And then execute:
-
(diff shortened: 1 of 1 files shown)


peterzhu2118 (changes requested):
While I'm all for refactoring/rewriting the README, I'm not a fan of deleting a large chunk of it. The README is the first document a user sees when they navigate to this repo, so it should describe what this library is, how to install it, and a quick start guide that highlights the most common ways to use this library (and perhaps linking to separate documentation for more advanced things).

BurdetteLamar:
> While I'm all for refactoring/rewriting the README, I'm not a fan of deleting a large chunk of it. The README is the first document a user sees when they navigate to this repo, so it should describe what this library is, how to install it, and a quick start guide that highlights the most common ways to use this library (and perhaps linking to separate documentation for more advanced things).

I'll be happy to restore the Usage section.  The Installation section should remain deleted?  Along with the remark about installation in the Development section?

peterzhu2118:
I think the Installation and Usage sections should be kept (or rewritten). I also think the blurb at the beginning should be expanded, IMO the one-liner is a bit too simple. The other parts I'm ok with deleting.

BurdetteLamar:
> I think the Installation and Usage sections should be kept (or rewritten). I also think the blurb at the beginning should be expanded, IMO the one-liner is a bit too simple. The other parts I'm ok with deleting.

Re Installation:  Doesn't Net::HTTP come with Ruby nowadays?  So no installation needed?

peterzhu2118:
Yes, it's bundled with Ruby, but it might not be the latest version, in which case they'll have to install it from RubyGems.

BurdetteLamar:
Stet, then.

PR #119, opened 2023-02-09
[DOC] Enhanced RDoc for Compression

BurdetteLamar:


BurdetteLamar:
In Net::HTTPResponse#inflater:
```
  def inflater # :nodoc:
    return yield @socket unless Net::HTTP::HAVE_ZLIB
    return yield @socket unless @decode_content
    return yield @socket if self['content-range']
```

- I think that these days ZLib is always there, so not mentioned.
- Not clear to me how the second condition works, so not (yet) mentioned.
- I think I got the third one right.

peterzhu2118 (approved):

PR #120, opened 2023-02-13
[DOC] Enhanced RDoc for request headers

BurdetteLamar:
Adds documentation to Net::HTTPRequest and add links to it from the request subclasses.


peterzhu2118 (commented):
I think this removes all usages of `:include: doc/net-http/included_setters.rdoc` (except in one place in request.rb), so can we delete that file and inline the contents?

BurdetteLamar:
> I think this removes all usages of `:include: doc/net-http/included_setters.rdoc` (except in one place in request.rb), so can we delete that file and inline the contents?

Done.  I've also added a header "Request Subclasses" to make a better transition.

peterzhu2118 (approved):

PR #121, opened 2023-02-14
[DOC] Enhanced RDoc for Net::HTTP#get

BurdetteLamar:
Actually trims most of the old doc, treating the method as what it is:  a wrapper.

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -1592,45 +1592,33 @@ module Net   #:nodoc:
 
     public
 
-    # Retrieves data from +path+ on the connected-to host which may be an
-    # absolute path String or a URI to extract the path from.
+    # :call-seq:
+    #    get(path, initheader = nil) {|res| ... }
     #
-    # +initheader+ must be a Hash like { 'Accept' => '*/*', ... },
-    # and it defaults to an empty hash.
-    # If +initheader+ doesn't have the key 'accept-encoding', then
-    # a value of "gzip;q=1.0,deflate;q=0.6,identity;q=0.3" is used,
-    # so that gzip compression is used in preference to deflate
-    # compression, which is used in preference to no compression.
-    # Ruby doesn't have libraries to support the compress (Lempel-Ziv)
-    # compression, so that is not supported.  The intent of this is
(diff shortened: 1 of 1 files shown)


lib/net/http.rb
```diff
@@ -1592,45 +1592,33 @@ def edit_path(path)
...
-    # so that gzip compression is used in preference to deflate
-    # compression, which is used in preference to no compression.
-    # Ruby doesn't have libraries to support the compress (Lempel-Ziv)
-    # compression, so that is not supported.  The intent of this is
-    # to reduce bandwidth by default.   If this routine sets up
-    # compression, then it does the decompression also, removing
-    # the header as well to prevent confusion.  Otherwise
-    # it leaves the body as it found it.
+    # :call-seq:
+    #    get(path, initheader = nil) {|res| ... }
```
peterzhu2118, line 1596: Why is this call-seq needed? And why is it different from the method definition?

lib/net/http.rb
```diff
@@ -1592,45 +1592,33 @@ def edit_path(path)
...
+    # Sends a GET request to the server.
+    # The request is based on the Net::HTTP::Get object
+    # created from string +path+ and initial headers hash +initheader+.
     #
-    # If called with a block, yields each fragment of the
-    # entity body in turn as a string as it is read from
-    # the socket.  Note that in this case, the returned response
-    # object will *not* contain a (meaningful) body.
+    # With a block given, returns a Net::HTTPResponse object,
+    # which actually will be an instance of a subclass of that class:
```
peterzhu2118, line 1603: The returned value doesn't seem to differ with or without a block? This paragraph seems to suggest that it does differ. Instead, we should talk about when the block gets called and the arguments it's called with.

lib/net/http.rb
BurdetteLamar: The old doc says not to use dest, so I call-seqed it without dest.

lib/net/http.rb
BurdetteLamar: Fixed (I think).

peterzhu2118 (approved):

PR #124, opened 2023-02-21
[DOC] Enhanced RDoc for Net:HTTP

BurdetteLamar:
Treats #request_get.

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -1855,6 +1855,7 @@ module Net   #:nodoc:
     #
     # With no block given, returns the response object (with the body already read):
     #
+    #   http = Net::HTTP.new(hostname)
     #   http.request_get('/todos') # => #<Net::HTTPOK 200 OK readbody=true>
     #
     # With a block given, calls the block with the response object
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -1090,7 +1090,7 @@ module Net   #:nodoc:
     end
 
     # Returns the continue timeout value.
-    # See Net::HTTP.continue_timeout=.
+    # See Net::HTTP.continue_timeout=. 
     #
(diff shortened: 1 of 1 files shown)


lib/net/http.rb
```diff
@@ -1090,7 +1090,7 @@ def write_timeout=(sec)
     end
 
     # Returns the continue timeout value.
-    # See Net::HTTP.continue_timeout=.
+    # See Net::HTTP.continue_timeout=.
```
peterzhu2118, line 1093: Please remove the trailing space.

```suggestion
    # See Net::HTTP.continue_timeout=.
```

lib/net/http.rb
```diff
@@ -1847,30 +1847,28 @@ def trace(path, initheader = nil)
...
+    #   http = Net::HTTP.new(hostname)
+    #   http.request_get('/todos') # => #<Net::HTTPOK 200 OK readbody=true>
     #
-    #     response = http.request_get('/index.html')
-    #     # The entity body is already read in this case.
-    #     p response['content-type']
-    #     puts response.body
+    # With a block given, calls the block with the response object
+    # (with the body _not_ already read),
+    # and returns the response object (_with_ the body already read):
```
peterzhu2118, line 1863: I don't think talking about whether the body has been read is important. It's an internal state that isn't publicly exposed (except in the output string of `Net::HTTPResponse#inspect`). So I'm not sure what the value of talking about it is.

lib/net/http.rb
BurdetteLamar: Fixed.

lib/net/http.rb
BurdetteLamar: Removed.

peterzhu2118 (approved):

PR #126, opened 2023-02-23

BurdetteLamar:
- :nodoc: for HTTP.initialize.
- :nodoc: for three constants that are undocumented, and that I think are not useful to the user.
- Adds/modifies doc for the first half (18/37) of the attributes.

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -926,7 +926,7 @@ module Net   #:nodoc:
       @debug_output = output
     end
 
-    # Returns the string host name or host IPA given as argument +address+ in ::new.
+    # Returns the string host name or host IP given as argument +address+ in ::new.
     attr_reader :address
 
     # Returns the integer port number given as argument +port+ in ::new.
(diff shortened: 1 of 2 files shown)


lib/net/http.rb
```diff
@@ -1103,14 +1118,18 @@ def continue_timeout=(sec)
...
+    # the still-open connection is used;
+    # otherwise the connection will have been closed
+    # and a new connection is opened.
     attr_accessor :keep_alive_timeout
 
-    # Whether to ignore EOF when reading response bodies with defined
-    # Content-Length headers. For backwards compatibility, the default is true.
+    # Sets or returns whether to ignore end-of-file when reading a response body
+    # with <tt>Content-Length</tt> headers;
+    # initially +true+
```
peterzhu2118, line 1132: Nit: period at the end of sentence.

```suggestion
    # initially +true+.
```

peterzhu2118 (approved):

PR #130, opened 2023-03-05
Enhanced RDoc for Net::HTTP

BurdetteLamar:
Treats a couple of methods;  most are private or nodoc.

### DIFF
--- a/lib/net/http/response.rb
+++ b/lib/net/http/response.rb
@@ -224,13 +224,28 @@ class Net::HTTPResponse
   # Accept-Encoding header from the user.
   attr_accessor :decode_content
 
-  # The encoding to use for the response body. If Encoding, use that encoding.
-  # If other true value, attempt to detect the appropriate encoding, and use
-  # that.
+  # Returns the value set by body_encoding=, or +false+ if none;
+  # see #body_encoding=.
   attr_reader :body_encoding
 
-  # Set the encoding to use for the response body.  If given a String, find
-  # the related Encoding.
+  # Sets the encoding that should be used when reading the body:
+  #
+  # - If the given value is an Encoding object, that encoding will be used.
+  # - Otherwise if the value is a string, the value of
+  #   {Encoding#find(value)}[https://docs.ruby-lang.org/en/master/Encoding.html#method-c-find]
+  #   will be used.
+  # - Otherwise an encoding will be deduced from the body itself.
+  #
+  # Examples:
+  #
+  #   res = Net::HTTP.get_response(hostname, '/')
+  #   # => #<Net::HTTPOK 200 OK readbody=true>
+  #   res.body_encoding # => false
+  #   res.body_encoding = Encoding::US_ASCII
+  #   res.body_encoding # => #<Encoding:US-ASCII>
+  #   res.body_encoding = 'US-ASCII'
+  #   res.body_encoding # => #<Encoding:US-ASCII>
+   #
   def body_encoding=(value)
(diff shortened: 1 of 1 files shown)


lib/net/http/response.rb
```diff
@@ -224,13 +224,28 @@ def initialize(httpv, code, msg)   #:nodoc: internal use only
...
+  #
+  # Examples:
+  #
+  #   res = Net::HTTP.get_response(hostname, '/')
+  #   # => #<Net::HTTPOK 200 OK readbody=true>
+  #   res.body_encoding # => false
+  #   res.body_encoding = Encoding::US_ASCII
+  #   res.body_encoding # => #<Encoding:US-ASCII>
+  #   res.body_encoding = 'US-ASCII'
+  #   res.body_encoding # => #<Encoding:US-ASCII>
```
peterzhu2118, line 247: This example doesn't really show what the purpose of this method is. It just show that it is a setter.

lib/net/http/response.rb
```diff
@@ -362,26 +377,30 @@ def read_body(dest = nil, &block)
...
+  # note that repeated calls for the unmodified body return a cached string:
   #
-  # Calling this method a second or subsequent time will return the
-  # string already read.
+  #   path = '/todos/1'
+  #   Net::HTTP.start(hostname) do |http|
+  #     res = http.get(path)
+  #     p res.body
+  #     p res.body.object_id
+  #     p res.body.object_id
```
peterzhu2118, line 388: I'm not a fan of examples that use the object ID, I think we should just get rid of it.

lib/net/http/response.rb
BurdetteLamar: Not sure what you mean.  Suggestion?

lib/net/http/response.rb
BurdetteLamar: Removed.

lib/net/http/response.rb
peterzhu2118: What I mean is that this example isn't showing the purpose of the `#body_encoding=`.

An example like this shows the purpose:

```ruby
http = Net::HTTP.new(hostname)
req = Net::HTTP::Get.new('/')

http.request(req) do |res|
  p res.body.encoding # => #<Encoding:ASCII-8BIT>
end

http.request(req) do |res|
  res.body_encoding = "UTF-8"
  p res.body.encoding # => #<Encoding:UTF-8>
end

Comment (BurdetteLamar) [reply lib/net/http/response.rb]: Substituted in, and thanks.

**Comment by BurdetteLamar** [comment]
@peterzhu2118, ready for you.

File: lib/net/http/response.rb [hunk 3]
```diff
@@ -224,13 +224,32 @@ def initialize(httpv, code, msg)   #:nodoc: internal use only
...
+  #   req = Net::HTTP::Get.new('/')
+  #
+  #   http.request(req) do |res|
+  #     p res.body.encoding # => #<Encoding:ASCII-8BIT>
+  #   end
+  #
+  #   http.request(req) do |res|
+  #     res.body_encoding = "UTF-8"
+  #     p res.body.encoding # => #<Encoding:UTF-8>
+  #   end  #   res.body_encoding # => false
```
peterzhu2118, line 251: Is the `# res.body_encoding # => false` on this line leftover?

BurdetteLamar: Yes.  Removed.

peterzhu2118 (approved):

PR #134, opened 2023-04-21
Update CI for bundler caching

ytkg:


ytkg:
Oops. Test failed.

hsbt:
The current configuration is expected behavior. CI time is short. There is no strong reason we should use bundler cache.

PR #145, opened 2023-08-13
Improve performance of HTTPHeader#content_type

technicalpickles:
In the existing implementation, `main_type` and `sub_type` would end up being called multiple times potentially.

Instead of doing that, save the result so it can be re-used.

### DIFF
--- a/lib/net/http/header.rb
+++ b/lib/net/http/header.rb
@@ -699,10 +699,14 @@ module Net::HTTPHeader
   #   res.content_type    # => "application/json"
   #
   def content_type
-    return nil unless main_type()
-    if sub_type()
-    then "#{main_type()}/#{sub_type()}"
-    else main_type()
+    main = main_type()
+    return nil unless main
+
+    sub = sub_type()
+    if sub
+      "#{main}/#{sub}"
+    else
+      main
     end
   end
 


technicalpickles:
Benchmark:
```
require 'benchmark/ips'
require 'net/http'

content_type = "application/json; charset=utf-8"

class Current
  include Net::HTTPHeader
  def initialize
    initialize_http_header({"content-type" => "application/json; charset=utf-8"})
  end
end

class Refactored
  include Net::HTTPHeader
  def initialize
    initialize_http_header({"content-type" => "application/json; charset=utf-8"})
  end

  def content_type
    main = main_type()
    return nil unless main

    sub = sub_type
    if sub
      "#{main}/#{sub}"
    else
      main
    end
  end
end

Benchmark.ips do |x|
  x.report "content_type current" do
    Current.new.content_type
  end

  x.report "content_type refactored" do
    Refactored.new.content_type
  end

  x.compare!
end
```

Result:

```
Warming up --------------------------------------
content_type current         27.752k i/100ms
content_type refactored   45.404k i/100ms
Calculating -------------------------------------
content_type current         281.205k (± 0.7%) i/s -      1.415M in   5.033402s
content_type refactored   452.947k (± 0.8%) i/s -      2.270M in   5.012417s

Comparison:
content_type refactored:   452946.6 i/s
content_type current:         281205.3 i/s - 1.61x  slower
```

yuki24:
Thank you!

Issue #143, opened 2023-07-07
Config headers to not be capitalized


AllanSiqueira:
## Context
I am consuming a API that have case sensitive header keys.
It is not common to see an case-sensitive header and also is off the [specification](https://developer.mozilla.org/en-US/docs/Glossary/HTTP_header). But sometimes it is off our control.

What I have in mind is some config to making possible to prevent the capitalize behavior and pass the header without changing it. Something like this:

```ruby
http = Net::HTTP.new(endpoint.host, endpoint.port)
http.preserve_headers = true
```

The default behavior will remain the same, the idea is only change the behavior if this config are true.

Some workaround I can do is making a monkeypath to prevent the capitalization, but I thing it make sense to have this configuration.

I can try to implement it, but I want to know if make sense for Net::HTTP.

technicalpickles:
I have been looking at how headers are capitalized for other reasons, and there is, fortunately, only one place where it is defined in `Net::HTTPHeader#capitalize`, and that module is included in `Net::HTTPResponse` and `Net::HTTPGenericRequest`.

I could imagine implementing this behavior by having some attribute in those classes which controls if `capitalize` is a no-op, and `http.preserve_headers == true` would handle setting that attribute.

AllanSiqueira:
Great, I will work on it!

AllanSiqueira:
@yuki24 Sorry to tag you, do you think this change make sense? Can I do something to clarify or make it better?

shayonj:
+1 on preserving headers, since its the default in other languages too and an attribute to control this with `preserve_headers` would be great

PR #146, opened 2023-08-16
Add new configuration to preserve headers case

AllanSiqueira:
Issue: #143
- Add new configuration to preserve headers case in Net::HTTP class
- Add new configuration to preserve headers case in GenericRequest class


nobu (changes requested):
The name doesn’t feel nice to me.
What of headers does it preserve?

AllanSiqueira:
> The name doesn’t feel nice to me. What of headers does it preserve?

Great, what do you think we call it `capitalize_headers`, make it default true.

AllanSiqueira:
Done the changes, when you have the time, may you review it again @nobu ?

PR #147, opened 2023-08-18
[Feature #19641] Enable to pass SSLContext to Net::HTTP

shouichi:
Also SSLContext is reused among requests.

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -1583,11 +1583,15 @@ module Net   #:nodoc:
     end
     private :do_start
 
+    def ssl_context=(context)
+      @user_provided_ssl_context = context
+    end
+
     def connect
       if use_ssl?
         # reference early to load OpenSSL before connecting,
         # as OpenSSL may take time to load.
-        @ssl_context = OpenSSL::SSL::SSLContext.new
+        @ssl_context ||= prepare_ssl_context(@user_provided_ssl_context)
       end
 
       if proxy? then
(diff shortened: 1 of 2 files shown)


rhenium:
Reusing SSLContext might be a breaking change. A Net::HTTP instance may start a connection multiple times, but an SSLContext can only be modified before the first connection due to `openssl`'s limitations for thread safety. An example which won't work if SSLContext is reused:

```ruby
http = Net::HTTP.new(host)
http.cert = cert_a
http.key = key_a
http.start {
  http.get("/")
}

http.cert = cert_b
http.key = key_b
http.start {
  http.get("/")
}
```

This concern led to the hack-ish code for TLS session resumption we currently have in #connect. Please see https://bugs.ruby-lang.org/issues/5341 for context.

rhenium:
https://bugs.ruby-lang.org/issues/19641
> A potential drawback is conflict handling may not be obvious to users. For example, what happens if a user both sets Net::HTTP#verify_hostname and OpenSSL::SSL::Context#verify_hostname?

For the same reason as above, the new attribute for setting SSLContext has to assume the user-supplied SSLContext may already be frozen. So I think we have no choice but to disallow other existing attributes for configuring SSLContext if user supplies a pre-made SSLContext.

shouichi:
@rhenium Thank you for the explanation. How about letting users set arbitrary parameters to SSLContext?

```
http.ssl_context_params = { security_level: 1}
```

PR #156, opened 2023-09-19
Add support for URI that contains an IPv6 address

### LINKED ISSUE #73: Incorrect Host header for IPv6 addresses
When specifying a URI with an IPv6 address as the host, the Host header of the request contains the IPv6 address but it is not surrounded in brackets

e.g. With a locally running apache webserver, the following snippet results in an HTTP 400 Bad Request error because of the malformed host header (`Host: ::1` instead of `Host: [::1]`)
```
require 'net/http'
uri = URI.parse('http://[::1]')
Net::HTTP.get_response(uri)
```

0x1eef:
Fix #73:
Net::HTTP.get_response(URI.parse("http://[::1]"))

When the 'Host' header does not wrap an IPv6 address
in brackets, some servers (eg Apache) reject the request
with a 400 status code.

### DIFF
--- a/lib/net/http/generic_request.rb
+++ b/lib/net/http/generic_request.rb
@@ -19,10 +19,10 @@ class Net::HTTPGenericRequest
 
     if URI === uri_or_path then
       raise ArgumentError, "not an HTTP URI" unless URI::HTTP === uri_or_path
-      hostname = uri_or_path.hostname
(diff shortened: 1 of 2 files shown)


0x1eef:
It doesn't look like this pull request will be merged. Closing.

PR #162, opened 2023-11-25
test/net/http/test_https.rb - fixup session reuse tests

MSP-Greg:
Add `test_session_reuse_with_timeout` test, which verifies that setting `ssl_timeout` will allow session reuse.  Another test checks that not setting `ssl_timeout` works, as it uses OpenSSL's default value.

Rename `test_session_reuse_but_expire` to `test_session_reuse_with_expired_timeout` and fixup for WIndows mingw/ucrt builds.

### DIFF
--- a/test/net/http/test_https.rb
+++ b/test/net/http/test_https.rb
@@ -164,7 +164,7 @@ class TestNetHTTPS < Test::Unit::TestCase
     http.finish
   end
 
-  def test_session_reuse_but_expire
+  def test_session_reuse_with_timeout
     # FIXME: The new_session_cb is known broken for clients in OpenSSL 1.1.0h.
     omit if OpenSSL::OPENSSL_LIBRARY_VERSION.include?('OpenSSL 1.1.0h')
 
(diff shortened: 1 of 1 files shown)


MSP-Greg:
Duplicate of https://github.com/ruby/ruby/pull/9033.  Not sure which is preferred.

rhenium:
I wonder why this is failing on Windows only, as it doesn't seem to be relying on platform-dependent features.

MSP-Greg:
@rhenium

Yesterday (Thanksgiving) was not a good day for finding a block of time to work on this.  It's obviously upstream of net-http.

I just noticed that MSYS2 has already jumped to OpenSSL 3.2.0.  Note that the only build 'fully' using it is the urcrt build, as head is still a bit of both (see [here](https://github.com/MSP-Greg/github-actions-ruby-info/actions/runs/6989487185/job/19018079568#step:4:60)).  I suspect MSFT/vcpkg will update soon also, but it's still using 3.1.4.

I'm going to add ucrt to #163 and see if it passes...

MSP-Greg:
I ran CI in my fork of [ruby/openssl](https://github.com/MSP-Greg/ruby-openssl/actions/runs/6989779586), and Windows ucrt passed.

Is this covered by the tests there?

PR #165, opened 2023-11-27
net/http.rb - fixup session timeout logic for OpenSSL 3

MSP-Greg:
In OpenSSL 3, `session.timeout` may not allow negative values, although `context.timeout` can be negative.  For more info see https://github.com/ruby/openssl/issues/703.

Since the code here is using `session.timeout` in a non-standard way, use `context.timeout`.  If it's nil, use `session.timeout`.  This allows the code to work with OpenSSL 3.2.0.

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -502,7 +502,7 @@ module Net   #:nodoc:
   # - {:ssl_timeout}[rdoc-ref:Net::HTTP#ssl_timeout]:
   #   Returns the ssl timeout.
   # - {:ssl_timeout=}[rdoc-ref:Net::HTTP#ssl_timeout=]:
-  #   Sets the ssl timeout.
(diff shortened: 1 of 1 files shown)


MSP-Greg:
I just rebased and additionally:
a) added Windows CI
b) removed the 'omit' in test_session_reuse_but_expire.

The branch is at https://github.com/MSP-Greg/net-http/commits/00-timeout-windows/, and [CI passed](https://github.com/MSP-Greg/net-http/actions/runs/8699525421/job/23858264214).

Windows Ruby head builds all use OpenSSL 3.2.1, Ruby 3.2 and 3.3 are using OpenSSL 3.2.0.

MSP-Greg:
Closing, see https://github.com/ruby/net-http/pull/178

PR #175, opened 2024-03-24
Add Net::HTTP.put method

Zopolis4:
As per title, as well as adding some missing documentation about `Net::HTTP#put` (and also adding documentation about the new `Net::HTTP.put` method)

Sanity tests pass locally: 
image

Running `rake test` failed for me, but that seemed to be a limitation of webrick? 

Not sure, which is why I am marking this PR as draft until I can find out where the fault lies.


Zopolis4:
Looked into it further-- webrick, at least in the way it is used here, does not appear to support `PUT`.

This became evident when creating the function `_test_request__PUT` (akin to `_test_request__POST`), which failed despite using `Net::HTTP::Put.new`.

hsbt:
Seems useful. Thanks.

PR #177, opened 2024-05-13
Add ability to configure default settings for new connections

### LINKED ISSUE #176: idea: global timeout settings
faraday allows setting global timeouts

```ruby
Faraday.default_connection_options = Faraday::ConnectionOptions.new({
  open_timeout: ENV.fetch('NET_HTTP_OPEN_TIMEOUT', 5).to_f,
  timeout: ENV.fetch('NET_HTTP_TIMEOUT', 30).to_f
})
```

If i set this, then i know all my own code which uses faraday, and gems that i use which use faraday (and don't set their own timeouts, which seems to usually be the case), will have this timeout set.

for net-http, i can do some combination of the bellow:

a) set it one by one for my own code
b) ask gem authors to allow this to be configured when it's not
c) monkeypatch gems
d) do a global monkeypatch of net-http
e) use the http-cage gem which has't been updated in 3 years (but maybe doesn't need to be, haven't really checked)

it would be great if net-http allowed a global setting, similar to faraday

i would be happy to submit a PR if folks think it's a good idea

fatkodima:
Closes #176.

It should be used by something like this:
```ruby
Net::HTTP.default_configuration = {
  read_timeout: 1,
  write_timeout: 1
}

http = Net::HTTP.new(hostname)
http.open_timeout   # => 60
http.read_timeout   # => 1
http.write_timeout  # => 1
```

Probably, there is a better name for the new option. And maybe we should allow more options to be configured, let me know.

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -1091,29 +1091,61 @@ module Net   #:nodoc:
       http
     end
 
+    class << HTTP
+      # Allows to set the default configuration that will be used
+      # when creating a new connection.
+      #
+      # Example:
+      #
+      #   Net::HTTP.default_configuration = {
+      #     read_timeout: 1,
+      #     write_timeout: 1
+      #   }
+      #   http = Net::HTTP.new(hostname)
+      #   http.open_timeout   # => 60
+      #   http.read_timeout   # => 1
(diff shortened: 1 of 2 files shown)


hsbt:
Thanks!

PR #178, opened 2024-06-21
test_https.rb - fix test_session_reuse_but_expire

MSP-Greg:
See issue https://github.com/ruby/openssl/issues/703

Currently, test code uses `http.ssl_timeout = -1`.  A negative timeout many would consider to be undefined.  As discussed in the ruby/openssl issue, OpenSSL 3.2 changed the timeout value type from signed to unsigned.

So, the previous test was relying on behavior that was undefined.

Changed the timeout value to `1`, and added a `sleep 1.25` command to wait for the second socket connection.

Using a negative value previously performed the equivalent of 'no session reuse'.  A better solution would be to add
a `session_cache_mode` accessor.  I can do that in another PR if there's interest.


hsbt:
I try this with `ruby/ruby`.

PR #189, opened 2024-11-15
Remove deprecated constants

nobu:
These constants, isolated in net/http/backward.rb, have not only been deprecated since 2001, but have also had a warning since 2021.


hsbt (approved):

yahonda:
I have opened https://bugs.ruby-lang.org/issues/20900 because I wanted `Net.send(:remove_const, :HTTPSession)` should emit warnings first.

PR #199, opened 2024-12-15
[DOC] Fix broken rdoc-ref links

st0012:
Found through https://github.com/ruby/rdoc/pull/1241

There are 2 changes:

1. Several methods' `aliased as` description are not linking properly because those alias methods have been marked as `:nodoc: obsolete` for a long time, which breaks the `rdoc-ref` links. Unless we want to encourage users to use them again, we don't need to mention them in the documentation.
2. We should link to `OpenSSL::SSL::SSLContext` methods using the proper `rdoc-ref` instead of linking to `docs.ruby-lang.org` to make sure their linked properly after synced to `ruby/ruby`.

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -1546,11 +1546,11 @@ module Net   #:nodoc:
     attr_accessor :cert_store
 
     # Sets or returns the available SSL ciphers.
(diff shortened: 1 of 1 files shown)


st0012:
cc @BurdetteLamar

jeremyevans (approved):

PR #206, opened 2025-02-03
Freeze some constants to improve Ractor compatibility

osyoyu:
Freeze `Net::HTTP::SSL_IVNAMES`, `Net::HTTPResponse::CODE_CLASS_TO_OBJ` and `Net::HTTPResponse::CODE_TO_OBJ` to improve Ractor compatibility.

This change allows the following code to work:

```ruby
Ractor.new {
  uri = URI.parse('http://example.com')
  http.open_timeout = nil
  http.read_timeout = nil
  http.get('/index.html')
}
```

---

Possibilities are that this patch breaks code which modify these constants. I have quickly skimmed over GitHub Code Search and have found only one instance where `CODE_TO_OBJ` is mutated, from 17-year-old code:
https://github.com/tolsen/rubydav/blob/62776abc293073da894b08c22d76952c918525c6/lib/rubydav/webdav.rb#L267

`rubydav` already needs a number of modifications to run under Ruby 3.4.1 even without this patch, so I think this change won't introduce additional major damage.

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -1529,7 +1529,7 @@ module Net   #:nodoc:
       :verify_hostname,
     ] # :nodoc:
 
-    SSL_IVNAMES = SSL_ATTRIBUTES.map { |a| "@#{a}".to_sym } # :nodoc:
+    SSL_IVNAMES = SSL_ATTRIBUTES.map { |a| "@#{a}".to_sym }.freeze # :nodoc:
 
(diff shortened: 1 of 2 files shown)


marshall-lee:
@hsbt Hi! Any chances this one is merged? Making HTTP requests in Ruby ractors become more & more desirable. And it seems like a safe change anyway. This sure will help a lot. Thanks

Issue #205, opened 2025-01-22
Do not supply a default content type.


mullermp:
Hello.

I would like to know if you would welcome a change to remove `[suppy_default_content_type](https://github.com/ruby/net-http/blob/ae6c43c2961fcceb877494a0f96ae3ad93203737/lib/net/http/generic_request.rb#L263)`.

This has caused issues with AWS services, where content type may be a modeled API parameter in REST services, like with S3 where you can specify the content type of an object. Currently we work around this with a patch:

```
Thread.current[:net_http_skip_default_content_type] = true

def self.apply!
  Net::HTTPGenericRequest.prepend(PatchDefaultContentType)
end

module PatchDefaultContentType
  def supply_default_content_type
    return if Thread.current[:net_http_skip_default_content_type]

    super
  end
end
```

Checking RFC for Content-Type: https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.5

```
   A sender that generates a message containing a payload body SHOULD
   generate a Content-Type header field in that message unless the
   intended media type of the enclosed representation is unknown to the
   sender.  If a Content-Type header field is not present, the recipient
   MAY either assume a media type of "application/octet-stream"
   ([RFC2046], Section 4.5.1) or examine the data to determine its type.
```

Based on the RFC, I think Net::HTTP behavior is incorrect - assuming a default content type is not correct because the media type is not known unless Net::HTTP reads/inspects the body. The receiver may assume octet-stream unless provided.

knu:
We agreed to fix this at the risk of breaking compatibility.

hsbt:
We try to remove default content type for next release. If we faced large issue with this change, we may revert this or apply another approach.

hsbt:
see also https://github.com/ruby/net-http/issues/239#issuecomment-3504754610

tomwrobel:
> Is there a place you are collecting reports of this causing problems for people? If so, what information would be useful to report?
> 
> I think this is an improvement, but many users are impacted where they depended on this behavior. Our organization has seen problems from this change.

This was a significant breaking change for us, and a VERY subtle bug to track down. I realise that the repository doesn't use semantic versioning, but it's quite a breaking change. I would have preferred adding warnings that this feature was going to change before it was made

hsbt:
You don’t know we already warn that https://github.com/ruby/net-http/pull/207/files#diff-1c8f56dc07513bad783e1dcf9353debd14fd86fbd79ad12fcc12ffaef9426573L378

[...]

PR #207, opened 2025-02-13
Don't set content type by default

hsbt:
Fixes https://github.com/ruby/net-http/issues/205


Resonious:
We had this change pulled in from a transitive dependency during a routine dependabot PR. Had no idea there was a breaking change because the main PR and changelog was for another library. 

Does this project not use semver? I thought breaking changes belonged in major version updates. And I've never heard of 0.6->0.7 being considered major - perhaps that's ignorance on my part!

PR #223, opened 2025-07-14
Replace Timeout.timeout with Socket.tcp(..., open_timeout:)

osyoyu:
Resolves https://github.com/ruby/net-http/issues/6.

This patch replaces the implementation of #open_timeout from Timeout.timeout from the builtin timeout in Socket.tcp, which was introduced in Ruby 3.5 (https://bugs.ruby-lang.org/issues/21347).

The builtin timeout in Socket.tcp is better in several ways. First, it does not rely on a separate Ruby Thread for monitoring Timeout (which is what the timeout library internally does). Also, it is compatible with Ractors, since it does not rely on Mutexes (which is also what the timeout library does).

This change allows the following code to work.

```ruby
require 'net/http'
Ractor.new {
  uri = URI('http://example.com/')
  http = Net::HTTP.new(uri.host, uri.port)
  http.open_timeout = 1
  http.get(uri.path)
}.value
```

In Ruby <3.5 environments where `Socket.tcp` does not have the `open_timeout` option, I have kept the behavior unchanged. net/http will use `Timeout.timeout { TCPSocket.open }`.

## Changes in behavior

The raised `Net::OpenTimeout`'s message has slightly changed, and also carrys a `Errno::ETIMEDOUT` as its cause. I can keep the message compatible if needed.

```
/home/osyoyu/Development/rubyorg/net-http/lib/net/http.rb:1668:in 'Net::HTTP#connect': Failed to open TCP connection to example.com:80 (Connection timed out - user specified timeout)
(Net::OpenTimeout)
        from /home/osyoyu/Development/rubyorg/net-http/lib/net/http.rb:1636:in 'Net::HTTP#do_start'
        from /home/osyoyu/Development/rubyorg/net-http/lib/net/http.rb:1625:in 'Net::HTTP#start'
        from /home/osyoyu/Development/rubyorg/net-http/lib/net/http.rb:2376:in 'Net::HTTP#request'
        from /home/osyoyu/Development/rubyorg/net-http/lib/net/http.rb:1991:in 'Net::HTTP#get'
        from nethttptest.rb:13:in '<main>'
/home/osyoyu/.rbenv/versions/master/lib/ruby/3.5.0+2/socket.rb:898:in 'block in Socket.tcp_with_fast_fallback': Connection timed out - user specified timeout (Errno::ETIMEDOUT)
        from /home/osyoyu/.rbenv/versions/master/lib/ruby/3.5.0+2/socket.rb:728:in 'Kernel#loop'
[...]

PR #224, opened 2025-07-16
Replace Timeout.timeout with TCPSocket.open(open_timeout:) when available

osyoyu:
This patch replaces the implementation of #open_timeout from Timeout.timeout from the builtin timeout in TCPSocket.open, which was introduced in Ruby 3.5 (https://bugs.ruby-lang.org/issues/21347).

The builtin timeout in TCPSocket.open is better in several ways. First, it does not rely on a separate Ruby Thread for monitoring Timeout (which is what the timeout library internally does). Also, it is compatible with Ractors, since it does not rely on Mutexes (which is also what the timeout library does).

This change allows the following code to work.

    require 'net/http'
    Ractor.new {
      uri = URI('http://example.com/')
      http.open_timeout = 1
      http.get(uri.path)
    }.value

In Ruby <3.5 environments where `TCPSocket.open` does not have the `open_timeout` option, I have kept the behavior unchanged. net/http will use `Timeout.timeout { TCPSocket.open }`.


lib/net/http.rb
```diff
@@ -1654,14 +1654,22 @@ def connect
...
-        begin
-          TCPSocket.open(conn_addr, conn_port, @local_host, @local_port)
-        rescue => e
-          raise e, "Failed to open TCP connection to " +
-            "#{conn_addr}:#{conn_port} (#{e.message})"
+      begin
+        s = begin
+          # Use built-in timeout in TCPSocket.open if available
+          TCPSocket.open(conn_addr, conn_port, @local_host, @local_port, open_timeout: @open_timeout)
+        rescue ArgumentError => e
```
osyoyu, line 1661: I couldn't come up with a better way to detect the absence of `open_timeout:` option. Is this a acceptable solution?

TCPSocket.open is basically `(**args)` from the perspective of Ruby, so Method#parameters wasn't an option:

```
irb(main):001> require 'socket'
=> true
irb(main):002> TCPSocket.method(:open).parameters
=> [[:rest]]
```

osyoyu: Actually it's not only very old Rubies, but even 3.4 raises on a `TCPSocket.open` call with `open_timeout`. It is true that the situation is different in Ruby 2.x, where keyword arguments were not a argument of its own kind (workaround in 09bf573dd5bd7111b56c57918c8a3a84b068309b).

deivid-rodriguez: Oh, I thought your question was specifically for the Ruby 2.x workaround, since you need to parse a very generic error message, but I see how the detection for Ruby 3.4 is also brittle since it also involves parsing the error message, even if more specific. Unfortunately, I don't know of a better way, but I'd say this way is acceptable.

mohamedhafez: Considering that every single new HTTP request is going to go through this raise/rescue flow on ruby < 3.5, and that exceptions are kind of expensive, personally I think a better option is just to have test for `RUBY_VERSION.to_f < 3.5` directly to see if we use the old `Timeout.timeout`. I get that testing version numbers directly is a bit distasteful, but when there's actual performance issues and brittleness issues on the line... I'd say its merited

osyoyu: I've took @eregon's idea. Pushed in 06d982f3a178a65777332ed402d5a80c4b3216da. (thanks: @niku)

osyoyu: Yes, that’s exactly right. Using class variables would unfortunately kill Ractors.

osyoyu:
Had a chance to chat with @shioimm regarding my previous comment and #223. She advised not to use `Socket.tcp` for net/http for performance reasons. (I don't have any numbers at hand, but maybe I can run some benchmarks later)

niku:
We've identified several options at this point. To make them easier to review, I've compiled them into a table.

| Method | Advantages (Pros) | Disadvantages (Cons) | Notes
| :--- | :--- | :--- | :--- 
| Check the result of `TCPSocket.method(:open).parameters` | **Most direct** | It's unknown/difficult to get a result more specific than `[[:rest]]` from a C extension. | https://github.com/ruby/net-http/pull/224#discussion_r2217063020 |
| Try calling `TCPSocket.open` with `open_timeout` and **fallback** on `ArgumentError` | The check (if it can run) and the actual execution method **are identical**. There are **no other considerations**. | **Requires code** to catch the `ArgumentError` and perform the fallback. |  this PR |
| **Assume** `TCPSocket.method(:open)` accepts `open_timeout` if `Socket.method(:tcp).parameters` does | Keeps code complexity **low**. | It's **not intuitive** why this workaround is used. / There is **no guarantee** this assumption is always true. | https://github.com/ruby/net-http/pull/224#issuecomment-3473072497 |
| Use `Socket.tcp` | Keeps code complexity **low**. / The check and the method to be executed **are identical**, with no other considerations. | **Poor performance**. | https://github.com/ruby/net-http/pull/224#issuecomment-3500000909 | 
| Check the Ruby version (`RUBY_VERSION`) | **Simple** | May **not work correctly on implementations other than CRuby**. / **Will fail** if `open_timeout` is removed for some reason in a future version (e.g., Ruby 3.5+). | https://github.com/ruby/net-http/pull/224#discussion_r2411676663 https://github.com/ruby/net-http/pull/224#discussion_r2500181341 |

At this point, this PR's changes seem to be the most suitable for me 👍

shioimm:
I also support the approach of calling `TCPSocket.open` with `open_timeout`, and raising an `ArgumentError` if it’s not available.
This makes the intention clear and ensures the behavior is independent of the Ruby implementation.
For these reasons, I will merge this PR.

osyoyu:
Thank you everyone for comments & reviews!

osyoyu:
Oh, definitely not expected. I'll take a look. I think there is no workaround besides downgrading net-http.

eregon:
> Is there any place where a changelog should be recorded? I can send a patch to CRuby’s NEWS if that’s the place.

I think just the release notes of this gem is fine (and that's automatic).

mohamedhafez:
So actually @osyoyu  and others: this same functionality would also be beneficial in net-smtp, and other net-* gems. What do you all think about extracting this into a gem that monkeypatches TCPSocket to provide a new method timeoutable_connect that encapsulates this new behavior, so we can implement it once and make improvements in one place, instead of having to recreate it in several different gems? Each net-* gem could just depend on the new gem we'd create, and use the new method it introduces

eregon:
@mohamedhafez IMO a gem just for that would be overkill and add a lot of complication for little duplication. But I see that as a sign that the current solution is maybe over-complicated and maybe we should simplify:

From https://github.com/ruby/net-http/pull/224#issuecomment-3505401550 I believe
> Assume TCPSocket.method(:open) accepts open_timeout if Socket.method(:tcp).parameters does

is actually very reasonable, because if `TCPSocket.open` accepts `open_timeout`, it's trivial to add it to `Socket.tcp`. If that doesn't hold it would be trivial to fix in that Ruby implementation, hence very unlikely to be a problem in practice.

The big advantage is that check can be stored in a constant and be done only once, vs having to try `TCPSocket.open(open_timeout: timeout)` for every Net::HTTP instance, which is both faster and simpler.

Is anyone against that approach? If not I can make a PR for it.

lib/net/http.rb
eregon: > `resolv_timeout` is extremely complex to handle, CRuby has had many iterations of it (some of them buggy, leaking, segfaulting and other issues)

Can't resist to link https://bugs.ruby-lang.org/issues/21790 which is another case of "it's very messy to do getaddrinfo with timeout i.e. resolv_timeout".

PR #227, opened 2025-09-04
[DOC] Fix links

BurdetteLamar:


jeremyevans (commented):
Looking more into this, I see the following in `net/http.rb`:

```ruby
  HTTPSession = HTTP
  deprecate_constant :HTTPSession
```

Not sure why the rdoc shows `Net::HTTPSession` on the page for `Net::HTTP` (https://docs.ruby-lang.org/en/master/Net/HTTP.html).

@hsbt @nurse Can we remove `Net::HTTPSession` now, or should we wait until after the release of Ruby 3.5? I think that would fix the issue.

PR #231, opened 2025-10-08
Allow a URL String for shorthand methods

mame:
In short, this allows `Net::HTTP.get("http://example.com")` instead of `Net::HTTP.get(URI("http://example.com"))`.

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -819,7 +819,7 @@ module Net   #:nodoc:
           return http.request_get(path, &block)
         }
       else
-        uri = uri_or_host
+        uri = URI(uri_or_host)
         headers = path_or_headers
         start(uri.hostname, uri.port,
               :use_ssl => uri.scheme == 'https') {|http|
@@ -830,7 +830,7 @@ module Net   #:nodoc:
 
     # Posts data to a host; returns a Net::HTTPResponse object.
     #
-    # Argument +url+ must be a URL;
+    # Argument +url+ must be a URI;
     # argument +data+ must be a string:
     #
     #   _uri = uri.dup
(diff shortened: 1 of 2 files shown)


lib/net/http.rb
```diff
@@ -830,7 +830,7 @@ def HTTP.get_response(uri_or_host, path_or_headers = nil, port = nil, &block)
 
     # Posts data to a host; returns a Net::HTTPResponse object.
     #
-    # Argument +url+ must be a URL;
+    # Argument +url+ must be a URI;
```
olleolleolle: Minor: perhaps now this wording (used in many places) can be expanded to include "either as an instance of String or URI".

osyoyu:
I really appreciate this feature. Passing a String to Net::HTTP.get generates a pretty much cryptic error today:

```
irb(main):001> require 'net/http'
=> true
irb(main):002> Net::HTTP.get_response('https://example.com')
/Users/osyoyu/.rbenv/versions/dev/lib/ruby/gems/4.1.0+1/gems/net-http-0.9.1/lib/net/http.rb:824:in 'Net::HTTP.get_response': undefined method 'hostname' for an instance of String (NoMethodError)

        start(uri.hostname, uri.port,
                 ^^^^^^^^^
        from (irb):2:in '<main>'
        from /Users/osyoyu/.rbenv/versions/dev/lib/ruby/gems/4.1.0+1/gems/irb-1.17.0/exe/irb:9:in '<top (required)>'
        from /Users/osyoyu/.rbenv/versions/dev/lib/ruby/4.1.0+1/rubygems.rb:305:in 'Kernel#load'
        from /Users/osyoyu/.rbenv/versions/dev/lib/ruby/4.1.0+1/rubygems.rb:305:in 'Gem.activate_and_load_bin_path'
        from /Users/osyoyu/.rbenv/versions/dev/bin/irb:25:in '<main>'
```

PR #240, opened 2025-11-09
Add pattern matching support to Net::HTTPResponse

baweaver:
This PR adds `deconstruct_keys` to `Net::HTTPResponse`, enabling pattern matching on HTTP response attributes.

(Moved from https://github.com/ruby/ruby/pull/15109)

### Usage

```ruby
response = Net::HTTP.get_response(uri)
case response
in code: '200', content_type: /json/
  JSON.parse(response.body)
in code: '404'
  handle_not_found
end
```

### Implementation

• Added `PATTERN_MATCHING_KEYS` constant defining valid keys: code, message, http_version, body, content_type
• Implemented `deconstruct_keys` following Ruby pattern matching best practices
• Avoids expensive operations by accessing `@body` directly rather than calling the body method
• Returns only requested keys when specified, all valid keys when keys is `nil`

### DIFF
--- a/.document
+++ b/.document
@@ -1,0 +1,5 @@
+BSDL
+COPYING
+README.md
+doc/
+lib/
--- a/.github/workflows/push_gem.yml
+++ b/.github/workflows/push_gem.yml
@@ -23,7 +23,7 @@ jobs:
 
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@f4a75cfd619ee5ce8d5b864b0d183aff3c69b55a # v2.13.1
+        uses: step-security/harden-runner@95d9a5deda9de15063e7595e9719c11c38c90ae2 # v2.13.2
         with:
           egress-policy: audit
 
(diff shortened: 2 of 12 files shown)


test/net/http/test_httpresponse.rb
```diff
@@ -744,6 +744,46 @@ def test_inspect_response
...
+    keys = res.deconstruct_keys([:code, :message])
+    assert_equal({code: '200', message: 'OK'}, keys)
+  end
+
+  def test_pattern_matching
+    res = Net::HTTPOK.new('1.1', '200', 'OK')
+    res['content-type'] = 'application/json'
+
+    begin
+      matched = instance_eval <<~RUBY, __FILE__, __LINE__ + 1
```
baweaver: This was a similar trick as to what we had used around pattern matching for the 2.6 to 2.7 push, so I'm repeating it here. The other methods should function even without syntax bridges like this.

PR #241, opened 2025-11-09
Add pattern matching support to Net::HTTPGenericRequest

baweaver:
This PR adds deconstruct_keys to Net::HTTPGenericRequest, enabling pattern matching on HTTP request attributes.

(Moved from https://github.com/ruby/ruby/pull/15111)

### Usage

```ruby
case request
in method: 'POST', path: %r{^/api/}
  handle_api_request(request)
in method: 'GET', path: '/'
  handle_root
end
```

### Implementation

• Added `PATTERN_MATCHING_KEYS` constant defining valid keys: method, path, uri, body, content_type
• Implemented `deconstruct_keys` following Ruby pattern matching best practices
• Returns only requested keys when specified, all valid keys when keys is `nil`

### DIFF
--- a/test/net/http/test_http_request.rb
+++ b/test/net/http/test_http_request.rb
@@ -74,6 +74,18 @@ class HTTPRequestTest < Test::Unit::TestCase
     assert_equal "/foo", req.path
     assert_equal "example.com", req['Host']
 
+    req = Net::HTTP::Get.new(URI("https://203.0.113.1/foo"))
+    assert_equal "/foo", req.path
+    assert_equal "203.0.113.1", req['Host']
+
+    req = Net::HTTP::Get.new(URI("https://203.0.113.1:8000/foo"))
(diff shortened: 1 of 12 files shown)


test/net/http/test_http_request.rb
```diff
@@ -89,5 +89,45 @@ def test_header_set
...
+    keys = req.deconstruct_keys([:method, :path])
+    assert_equal({method: 'GET', path: '/test'}, keys)
+  end
+
+  def test_pattern_matching
+    uri = URI('https://example.com/api/users')
+    req = Net::HTTP::Post.new(uri)
+
+    begin
+      matched = instance_eval <<~RUBY, __FILE__, __LINE__ + 1
```
baweaver: This was a similar trick as to what we had used around pattern matching for the 2.6 to 2.7 push, so I'm repeating it here. The other methods should function even without syntax bridges like this.

PR #243, opened 2025-11-09
Add pattern matching support to Net::HTTPExceptions

baweaver:
This PR adds `deconstruct_keys` to the `Net::HTTPExceptions` module, enabling pattern matching on HTTP exception classes for cleaner error handling.

### Usage

```ruby
begin
  http.request(req)
rescue => e
  case e
  in HTTPRetriableError[response: res] if res.code == '503'
    retry_with_backoff
  in HTTPClientException[message: /not found/]
    handle_not_found
  in HTTPFatalError[response:]
    log_fatal_error(response)
  end
end
```

### Implementation

* Added `PATTERN_MATCHING_KEYS` constant to `HTTPExceptions` module with keys: message, response
* Implemented `deconstruct_keys` following Ruby pattern matching best practices
* Applies to all exception classes: `HTTPError`, `HTTPRetriableError`, `HTTPClientException`, `HTTPFatalError`
* Tests wrapped in `instance_eval` with `SyntaxError` rescue for Ruby 2.6 compatibility

### DIFF
--- a/lib/net/http/exceptions.rb
+++ b/lib/net/http/exceptions.rb
@@ -37,6 +37,7 @@ module Net
     end
   end
 
+  # :stopdoc:
   class HTTPError < ProtocolError
     include HTTPExceptions
   end
--- a/.document
+++ b/.document
@@ -1,0 +1,5 @@
+BSDL
+COPYING
+README.md
+doc/
+lib/
--- a/.github/workflows/push_gem.yml
+++ b/.github/workflows/push_gem.yml
@@ -23,7 +23,7 @@ jobs:
 
     steps:
       - name: Harden Runner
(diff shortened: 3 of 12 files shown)


lib/net/http/exceptions.rb
```diff
@@ -4,12 +4,37 @@ module Net
...
+    #
+    # Valid keys are: +:message+, +:response+
+    #
+    # Example:
+    #
+    #   begin
+    #     http.request(req)
+    #   rescue => e
+    #     case e
+    #     in HTTPRetriableError[response: { code: '503' }]
```
baweaver: Assuming that https://github.com/ruby/net-http/pull/240 is merged this is possible, otherwise nested deconstruction is not _quite_ possible hence the deviation in the tests.

Issue #251, opened 2025-12-08
v0.8.0 is not working with JRuby


hsbt:
From https://github.com/ruby/rubygems/pull/9089#issuecomment-3544818648

JRuby 10.0.2.0 and net-http 0.7.0 is working with some failures:

```
$ rake test
...
180 tests, 835 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
```

But net-http 0.8.0 is stucking while `rake test`.

I found this is caued by https://github.com/ruby/net-http/pull/224.

It means rubygems can't update vendored library in the future.

@shioimm @osyoyu Can you look that?

PR #252, opened 2025-12-08
Check whether TCPSocket#initialize supports open_timeout once and without exceptions

eregon:
* See discussion in https://github.com/ruby/net-http/pull/224
* This check is known to work on at least CRuby, TruffleRuby and JRuby.
* Exceptions show up with `ruby -d`/`$DEBUG == true` and would show for every Net::HTTP instance.
* Fixes https://github.com/ruby/net-http/issues/251

cc @osyoyu

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,7 @@ jobs:
     needs: ruby-versions
     name: test (${{ matrix.ruby }} / ${{ matrix.os }})
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


osyoyu:
For the hanging issue (#251), I also believe that there is something to be fixed on the JRuby side.
Given that, I'm not against this change, as mentioned in https://github.com/ruby/net-http/pull/224#issuecomment-3624092599.

For net-protocol, I agree with @eregon's standpoints in https://github.com/ruby/net-http/pull/252#issuecomment-3627758442. If things are fixable within net-http, it'd be better to do so, at least as the first action.

PR #255, opened 2025-12-13
Refactor HTTPS tests

rhenium:
This contains various improvements in tests for `openssl` integration:

  - Remove DHE parameters from test servers. OpenSSL is almost always compiled with ECC support nowadays and will prefer ECDHE over DHE.
  - Remove an outdated omission for a bug in OpenSSL 1.1.0h released in 2018. None of our CI systems use this specific OpenSSL version.
  - Use top-level `return` to skip tests if `openssl` is unavailable.
  - Refactor tests for `Net::HTTP#verify_callback`.


hsbt (approved):
👍

PR #279, opened 2026-03-06
Add RBS signatures from ruby/rbs repo

sferik:
See https://github.com/ruby/net-http/pull/278 for context.

### DIFF
--- a/sig/manifest.yaml
+++ b/sig/manifest.yaml
@@ -1,0 +1,3 @@
+dependencies:
+  - name: net-protocol
+  - name: uri
--- a/.github/workflows/push_gem.yml
+++ b/.github/workflows/push_gem.yml
@@ -34,6 +34,8 @@ jobs:
         with:
           bundler-cache: true
           ruby-version: ruby
+        env:
+          BUNDLE_WITHOUT: sig
 
       - name: Publish to RubyGems
         uses: rubygems/release-gem@1c162a739e8b4cb21a676e97b087e8268d8fc40b # v1.1.2
--- a/.github/workflows/sig.yml
+++ b/.github/workflows/sig.yml
@@ -1,0 +1,18 @@
+name: sig
+
+on: [push, pull_request]
+
+jobs:
+  sig:
+    runs-on: ubuntu-latest
+    env:
+      BUNDLE_WITH: sig
+    steps:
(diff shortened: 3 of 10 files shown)


ksss (commented):
The test code differs significantly from the [ruby/rbs version](https://github.com/ruby/rbs/blob/23daeea3f8075170788b24daba0cddd51488c54f/test/stdlib/Net_HTTP_test.rb), making it difficult to pinpoint the differences.
Can it be made as close to the original as possible?
If there are issues with the original, please let me know.

ksss (commented):
`NetHTTPTestSigHelper` and `NetHTTPRBSTestCase` should be removed.
This approach does not allow us to confirm if the `manifest.yaml` specification is sufficient.

Additionally, `library` should be used.
This is a DSL that explicitly states that only the specified library has been loaded.

```rb
  library "net-http"
  testing "singleton(::Net::HTTP)"
```

sig/manifest.yaml
```diff
@@ -0,0 +1,9 @@
+dependencies:
+  - name: net-protocol
+  - name: open-uri
+  - name: openssl
+  - name: socket
+  - name: strscan
+  - name: timeout
+  - name: uri
+  - name: zlib
```
ksss: This is a list of libraries required on the Ruby side, and many of them are not necessary as RBS type names.
If you haven't added types to RBS, there should be no additional type names.
Please copy it as is.

```suggestion
dependencies:
  - name: net-protocol
  - name: uri
```

ksss (commented):
I think it's almost good.
The `test_sig` should be excluded from `spec.files` in `net-http.gemspec` at the end..

ksss (commented):
LGTM

sferik:
@ksss Thank you for taking the time to thoroughly review it.

sferik:
Thank you very much! Did you discuss https://github.com/ruby/net-http/pull/280 and https://github.com/ruby/net-http/pull/281 at your dev meeting? I just rebased those from the `master` branch.

hsbt:
@sferik Sorry my late response.

We briefly discussed #280 and #281 at the dev meeting. The direction looks fine, but since the main code has changed, we want to make sure there are no regressions before merging. That's as far as the discussion went.

PR #303, opened 2026-06-21
Expose OpenSSL::SSL::SSLContext object

### LINKED ISSUE #139: Net::HTTP doesn't allow to set SSL options
### Context

Since OpenSSL 3.x, when a server close the TCP connection without first calling `SSL_shutdown`, the SSL client now error with `SSL_read: unexpected eof while reading`.

In OpenSSL 1.x, the client wouldn't mind and would behave like if the connection was cleanly closed.

To restore the 1.x behavior, you can set a specific options:

```ruby
ssl_context = OpenSSL::SSL::SSLContext.new
ssl_context.options |= OpenSSL::SSL::OP_IGNORE_UNEXPECTED_EOF
```

### Problem

The issue is that `Net::HTTP` only allow to set specific fields on the `SSLContext` object, and `options` is not one of them.

### Workaround

The issue can be worked around by changing the default options globally:

```ruby
if OpenSSL::SSL.const_defined?(:OP_IGNORE_UNEXPECTED_EOF)
  OpenSSL::SSL::SSLContext::DEFAULT_PARAMS[:options] |= OpenSSL::SSL::OP_IGNORE_UNEXPECTED_EOF
end
```

However this impact all SSL connections, not just the ones that need it.

### Solution

Not sure what the best API would be. But it would be great if we could directly pass a `SSLContext` instance to `Net::HTTP`, so that we're not limited on the SSL configuration.

rhenium:
Add `Net::HTTP#ssl_context` to return the `SSLContext` used for HTTPS connections, and reuse the same object for future connections made by the same `Net::HTTP` instance.

Also change existing SSL/TLS-related accessors to interact with the `SSLContext` directly, rather than temporarily storing values in instance variables and applying them later in `Net::HTTP#start`.

Note that the `SSLContext` becomes frozen after the first connection is made using the `Net::HTTP` instance. Although I do not expect this to affect most users, it is a compatibility concern.

[Feature #19641] https://bugs.ruby-lang.org/issues/19641

---

/cc @shouichi

An alternative to https://github.com/ruby/net-http/pull/147. Instead of accepting user-supplied `SSLContext`, this PR makes `Net::HTTP` prepare `SSLContext` earlier and expose it. This avoids ambiguity when both `Net::HTTP` accessors and `SSLContext` accessors are used.

This doesn't address my comment in the PR https://github.com/ruby/net-http/pull/147#issuecomment-1685201004, but I think this incompatibility may be unavoidable. It should be documented in the release note.

This resolves https://github.com/ruby/net-http/issues/139 by allowing users to do the following:

```ruby
http = Net::HTTP.new(host, 443)
http.use_ssl = true
http.ssl_context.options |= OpenSSL::SSL::OP_IGNORE_UNEXPECTED_EOF
```


shouichi:
Thank you for the work. Feel free to close #147.

PR #309, opened 2026-07-06
Add support for HTTP `QUERY` method

### LINKED ISSUE #202: Please add support for the QUERY HTTP method
On 2025-01-07 the [QUERY HTTP method, "a safe, idempotent request method that can carry request content." has reached the level of a Proposed Standard](https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/).

It solves the problem of making complex queries where the GET method is sometimes sufficient as it (according to the standard) lacks support for sending the query in the body, leaving only the query part of the URL which is limited. An often used workaround of using the POST method has a downside in that it's just the wrong method for this purpose as queries do not cause side-effects and this limits the retryabilty and cacheability of such queries.

Therefore It would be great to implement support for QUERY in Net::HTTP.

If you are open to it, I am interested in providing a PR.

seanpdoyle:
As of June 2026, [RFC 10008][] is a proposed standard.

From the RFC Abstract:

> This specification defines the QUERY method for HTTP. A QUERY requests
> that the request target process the enclosed content in a safe and
> idempotent manner and then respond with the result of that processing.
> This is similar to POST requests, but QUERY requests can be
> automatically repeated or restarted without concern for partial state
> changes.

This commit proposes support for the HTTP Query method by way of the `Net::HTTP::Query` class along with the corresponding `#query` and `#request_query` methods.

[RFC 10008]: https://www.rfc-editor.org/rfc/rfc10008.html

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -82,6 +82,7 @@ module Net   #:nodoc:
   #       body = 'Some text'
   #       http.post(path, body)  # Can also have a block.
   #       http.put(path, body)
+  #       http.query(path, body) # Can also have a block.
   #       http.delete(path)
   #       http.options(path)
   #       http.trace(path)
@@ -207,6 +208,7 @@ module Net   #:nodoc:
   #   - #options: OPTIONS.
   #   - #trace: TRACE.
   #   - #patch: PATCH.
+  #   - #query, #request_query: QUERY.
   #
   # - {WebDAV methods}[https://en.wikipedia.org/wiki/WebDAV#Implementation]:
   #
@@ -553,6 +555,8 @@ module Net   #:nodoc:
   #   Sends a PROPPATCH request and returns a response object.
   # - {#put}[rdoc-ref:Net::HTTP#put]:
   #   Sends a PUT request and returns a response object.
+  # - {#query}[rdoc-ref:Net::HTTP#query]:
+  #   Sends a QUERY request and returns a response object.
   # - {#request}[rdoc-ref:Net::HTTP#request]:
   #   Sends a request and returns a response object.
   # - {#request_get}[rdoc-ref:Net::HTTP#request_get]:
@@ -2101,6 +2105,25 @@ module Net   #:nodoc:
       send_entity(path, data, initheader, dest, Patch, &block)
     end
 
+    # Sends a QUERY request to the server;
+    # returns an instance of a subclass of Net::HTTPResponse.
+    #
+    # The request is based on the Net::HTTP::Query object
+    # created from string +path+, string +data+, and initial headers hash +initheader+.
+    #
+    #   data = '{"userId": 1, "id": 1, "title": "delectus aut autem", "completed": false}'
+    #   http = Net::HTTP.new(hostname)
+    #   http.query('/todos/1', data) # => #<Net::HTTPOK 200 OK readbody=true>
+    #
+    # Related:
+    #
+    # - Net::HTTP::Query: request class for \HTTP method QUERY.
+    # - Net::HTTP.query: sends QUERY request, returns response body.
+    #
+    def query(path, data, initheader = nil, dest = nil, &block) # :yield: +body_segment+
+      send_entity(path, data, initheader, dest, Query, &block)
+    end
+
     # Sends a PUT request to the server;
     # returns an instance of a subclass of Net::HTTPResponse.
     #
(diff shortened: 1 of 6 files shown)


lib/net/http/requests.rb
```diff
@@ -35,6 +35,41 @@ class Net::HTTP::Get < Net::HTTPRequest
   RESPONSE_HAS_BODY = true
 end
 
+# \Class for representing
```
rhenium: Between `Get` and `Head` is probably not the right place for this. Maybe this should go after `Patch`?

lib/net/http.rb
```diff
@@ -2120,6 +2120,25 @@ def put(path, data, initheader = nil)
...
+    #   body = '{"userId": 1, "id": 1, "title": "delectus aut autem", "completed": false}'
+    #   http = Net::HTTP.new(hostname)
+    #   http.query('/todos/1', body) # => #<Net::HTTPOK 200 OK readbody=true>
+    #
+    # Related:
+    #
+    # - Net::HTTP::Query: request class for \HTTP method QUERY.
+    # - Net::HTTP.query: sends QUERY request, returns response body.
+    #
+    def query(path, body, initheader = nil)
```
rhenium: This should take a block, like `#post`, since a response body is expected.

I wonder if `put` should have also taken a block, but that would be out of scope of this PR.

test/net/http/test_http.rb
```diff
@@ -665,6 +666,22 @@ def _test_request__GET(http)
     }
   end
 
+  def _test_request__QUERY(http)
+    req = Net::HTTP::Query.new('/')
+    http.request(req) {|res|
```
rhenium: I think a test with a request body is useful.

lib/net/http.rb
```diff
@@ -2335,6 +2354,36 @@ def request_put(path, data, initheader = nil, &block)   #:nodoc:
...
+    #
+    #   http.request_query('/todos') do |res|
+    #     p res
+    #   end # => #<Net::HTTPOK 200 OK readbody=true>
+    #
+    # Output:
+    #
+    #   #<Net::HTTPOK 200 OK readbody=false>
+    #
+    def request_query(path, body, initheader = nil, &block)   #:nodoc:
```
rhenium: Does anyone know why `request_put` is undocumented (marked as `:nodoc:`)? This appears to have been added in a 2003 commit, but I couldn't find the context: https://github.com/ruby/net-http/commit/38840ac04b4c953f4583a3eefdeb1522301b5c57

We also don't have  `request_patch` or `request_delete` corresponding to PATCH and DELETE methods.

Are `request_<method>` methods considered obsolete?

rhenium (commented):
Could you add it to `IDEMPOTENT_METHODS_` so that requests can be retried?

Class-level rdoc comments mention request method-specific methods. They should also updated to include the new methods for QUERY.

rhenium:
This will close https://github.com/ruby/net-http/issues/202

lib/net/http/requests.rb
seanpdoyle: I've moved this class definition below `Patch`.

lib/net/http.rb
seanpdoyle: I've modified the signature to reflect support for a block that yields the `body_segment` variable.

test/net/http/test_http.rb
seanpdoyle: I've added a `data` argument as a second request to both of this PR's new `http.request(req, …)` calls.

lib/net/http.rb
seanpdoyle: > [I wonder if put should have also taken a block](https://github.com/ruby/net-http/pull/309#discussion_r3530297827), but that would be out of scope of this PR.

Could they all be included in the diff, or would it be best to open two separate PRs:

1. to remove the `:nodoc:` for `#request_put` and to add support for passing a block to `#put`
2. to introduce `request_patch` and `request_delete`

I'm happy to open a separate PR to make these changes based on the response to `request_`-prefixed methods being deprecated.

rhenium: Changes to methods unrelated to the QUERY method should be in separate PRs.

Regarding those `request_<method>` methods, it's unclear to me if they're deprecated and we should avoid adding new ones, or if this was an oversight and methods for PATCH, DELETE, etc. are just missing. Personally I've never used them and they seem redundant, so I suspect it's the former, but I'm not sure.

I've requested a review from @nurse (the maintainer of net/http).

rhenium (approved):
Changes look good to me, apart from the open question about `request_query`.

PR #325, opened 2026-08-25
Limit read length of a single response line

### LINKED ISSUE #315: Ineffective request header size limit in #219
We received a vulnerability report that the request header size limit is not truly effective. This reproducer still shows unbound memory growth:
```ruby
require "socket"
require "net/http"
Thread.new do
  s = TCPServer.new("127.0.0.1", 18080)
  c = s.accept
  c.write("HTTP/1.1 200 ") # deliberately missing newline
  loop { c.write("A" * 16_384) } # continuous stream, no terminator
end
sleep 0.1
Net::HTTP.get(URI("http://127.0.0.1:18080/"))
```

As #219 was not treated as a vulnerability, I'm filing this issue publicly, too.

Report: [AISLE-Report-PSIRTSUPT-15870.md](https://github.com/user-attachments/files/29969227/AISLE-Report-PSIRTSUPT-15870.md)

hsbt:
A server that never sends a line terminator can grow the client's read buffer without bound, as reported in #315. The 1 MiB response header cap added by #219 only applies after a complete line has arrived, so it does not help against an unterminated status line, header line, chunk-size line or chunk trailer.

This PR reads all of those lines through the new `limit:` option of `Net::BufferedIO#readuntil` and fails with `Net::HTTPBadResponse` once a single line exceeds `MAX_RESPONSE_HEADER_LENGTH`. When the loaded net-protocol does not provide the option yet, the previous behavior is kept and the new tests are omitted, so this can be merged ahead of the net-protocol release. The net-protocol side is submitted separately.

Fixes #315


lib/net/http/response.rb
```diff
@@ -155,11 +155,24 @@ def read_new(sock)   #:nodoc: internal use only
       res
     end
 
+    if defined?(Net::ReadLimitExceeded)
```
rhenium: Should we (re-)add `net-protocol` dependency to the gemspec instead?

hsbt: Do you mean we specify supported version of `net-protocol`?

rhenium: Yes, the version that will include https://github.com/ruby/net-protocol/pull/67.

My understanding is that it had to be removed from `net-http.gemspec` in 2022 in commit 79d90c8154576c24b0fec201a711d17ab416c059 because of the conflict with RubyGems' use of stdlib net-http/net-protocol. That may no longer be an issue now that RubyGems vendors them.

hsbt: 🆗  Thanks for your explanation because I forgot that. I will do that.

PR #334, opened 2026-08-27
Match the 100-continue expectation case-insensitively

OskarEichler:
## Summary

Match the 100-continue expectation case-insensitively, using casecmp? on each existing header value. RFC 9110 section 10.1.1 specifies case-insensitive expectation values.

## Reproduction

Provide Expect: 100-CONTINUE to a POST with a continue timeout and a socket ready to return an early 417. Before, wait_for_continue skips the readiness check and early response. After, it behaves identically to lower-case 100-continue.

## Verification

- Ruby 4.0.6 through rbenv; existing `bundle exec rake test`: **201 tests, 914 assertions, zero failures/errors**, both baseline and this isolated patch.
- 28 focused checks cover lower/mixed/upper case, multiple header values, absent values, nonmatching prefixes/suffixes and HTTP/1.0 versus HTTP/1.1.
- Ruby syntax and `git diff --check` pass. Supplemental Lint scan retains the same 25 pre-existing findings; no new findings by file/cop.
- No test/spec files were added or modified, per this contribution's explicit no-new-tests constraint. Focused reproductions were run externally.

## Compatibility and limitations

No API or dependency change. Mixed-case 100-continue now engages the same wait behavior as lower-case. Matching remains exact per header field; comma-list parsing and quoted parameters are not introduced.

Based on master `23e859e92e10e43cad520fcacac3fd43640dba7b`. Optional external OpenSSL tests and other Ruby/OS combinations were not run. Existing related issues/PRs were checked; this is a focused correctness change, not a claim of exhaustive coverage.

Reference: https://www.rfc-editor.org/rfc/rfc9110.html#section-10.1.1


OskarEichler:
Superseded by #341, which preserves the case-insensitive match from this patch and also handles comma-separated Expect field values without substring false positives. Closing this narrower version so there is one canonical implementation to review.

